PR #2, opened 2019-10-08
v0.1.0

dpaez:
This PR adds:
- some code/repo organisation and tooling
- new constructor and init method
- WIP readme
- basic `example.js` file (it will be growing over time) showing some simple use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ publication](https://doi.org/10.3390/publications6020021) and this
 [technical publication](https://chartgerink.github.io/2018dat-com/)
 (note these might have been extended by now).
 
+:warning: **Work In Progress**
+
 ## Install
 
 `npm install @p2pcommons/sdk-js`
@@ -23,14 +25,11 @@ outside of it. If you are looking for a Command Line Interface (CLI),
 ## Example
 
 ```javascript
-const libsdk = require('@p2pcommons/sdk-js') // liberate science constructor function
-
-const content = libsdk({ type: 'content' })
-const profile = libsdk({ type: 'profile' })
+const libsdk = require('@p2pcommons/sdk-js')({}) // liberate science constructor function
 
 ;(async () => {
-  await module.init() // ~/.p2pcommons/hash/dat.json --> type: content
-  await profile.init() // ~/.p2pcommons/hash/dat.json --> type: profile
+  await libsdk.init({ type: 'content' }) // ~/.p2pcommons/hash/dat.json --> type: content
+  await libsdk.init({ type: 'profile' }) // ~/.p2pcommons/hash/dat.json --> type: profile
 })()
 ```
 
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ publication](https://doi.org/10.3390/publications6020021) and this
 [technical publication](https://chartgerink.github.io/2018dat-com/)
 (note these might have been extended by now).
 
+## Install
+
+`npm install @p2pcommons/sdk-js`
+
 ## Usage
 
 This is an `npm` module that does not export any processes to your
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -16,9 +22,21 @@ your NodeJS packages, but does not provide any direct functionality
 outside of it. If you are looking for a Command Line Interface (CLI),
 [we got you covered](https://github.com/libscie/cli).
```
chartgerink: Feel free to delete the CLI reference.

index.js
```diff
@@ -1,231 +1,132 @@
...
+    this.disableSwarm = !!opts.disableSwarm
+    if (!this.disableSwarm) {
+      debug('libsdk:constructor starting swarm')
+      this.swarm = discovery(Object.assign({}, DEFAULT_SWARM_OPTS, opts.swarm))
     }
+  }
+
+  async init ({ type, title = '', description = '', ...rest }) {
+    // follow module spec: https://github.com/p2pcommons/specs/pull/1/files?short_path=2d471ef#diff-2d471ef4e3a452b579a3367eb33ccfb9
+    // 1. create folder with name uniqueString()?
```
chartgerink: This should be the key of the Dat archive in the end, so uniqueString() would need to be a temporary storage only.

index.js
```diff
@@ -1,231 +1,132 @@
...
+    debug('libsdk:hash', hash)
+    if (!this.disableSwarm) {
+      this.swarm.add(archive)
+
+      archive.once('close', () => {
+        debug(`libsdk:closing archive ${archive.publicKey}...`)
+        const discoveryKey = DatEncoding.encode(archive.discoveryKey)
+        this.swarm.leave(discoveryKey)
+        this.swarm._replicatingFeeds.delete(discoveryKey)
+        // Note(dk): should we handle multiple drives?
```
chartgerink: I'm not sure I understand the question? One Dat archive should only be able to have one hyperdrive, if I understood the protocol correctly.

index.js
```diff
@@ -1,231 +1,132 @@
...
-    // when spawning a non-versioned module
-    // it automatically registers
-    // then adds as parent
-    // make sure to allow multiple parents ie multiSelect
-    // https://github.com/terkelg/prompts#autocompletemultiselectsame
-}
-////////////////////////////////////////////////////////////////////////////////
-
-// utils
+    datJSON.url = `dat://${hash}`
```
chartgerink: Do you think it's a good idea to drop the `dat://` to save some bytes?

index.js
```diff
@@ -1,231 +1,132 @@
...
 
-async function readMeta (hash, env) {
-    let file = path.join(env, hash, 'dat.json')
-    let meta = await fs.readFile(file)
+    // write dat.json
+    await writeFile(join(tmp, 'dat.json'), JSON.stringify(datJSON))
 
-    return JSON.parse(meta)
-}
+    // Note(dk): not necessary anymore?
```
chartgerink: If we have a database that indexes it we don't need this, yea :)

index.old.js
```diff
@@ -0,0 +1,238 @@
+const path = require('path')
+
+// old stack
+// const Dat = require('dat-node')
+// new stack
+// const hyperdrive = require('hyperpromise')
+// const storage = require('universal-storage')
```
chartgerink: Feel free to delete this file if you feel that it's redundant.

README.md
```diff
@@ -8,6 +8,12 @@ publication](https://doi.org/10.3390/publications6020021) and this
 [technical publication](https://chartgerink.github.io/2018dat-com/)
 (note these might have been extended by now).
 
+:warning: **Work In Progress**
+
+## Install
+
+`npm install @p2pcommons/sdk-js`
```
chartgerink: Could you please verifiy whether this fits with the updates to the `package.json` file?

index.js
dpaez: it is working as a temporary storage indeed

index.js
dpaez: Yup, you are right. I leave that as open question/reminder if we may manage multiple archives from the sdk. It is not a concern now.

index.js
dpaez: we can drop it. We will always see the url as the public key and most tools add the dat://.

index.js
dpaez: yup, it is coming. I guess I would end using level with some query mechanism and I would also like to use avro or protocol buffers not only for fast encode/decode but also the schema validation. That could be nice to have for content and profile modules.

index.old.js
dpaez: OK, it is handy to look when I need more background/content on some feature. It will be removed soon.

README.md
dpaez: @chartgerink current package.json name is `libscie-sdk`. Would you like to maintain that or use a different name? I guess I suggested the name in the readme but it's only that, a suggestion. 
Also, should we use a scope like `@p2pcommons`? This decision could affect all the packages inside this github org (consistency).

PR #25, opened 2019-11-01
Ongoing: v0.3.0

dpaez:
This is an ongoing PR adding `register` and `verify` methods. Feedback is welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,7 @@ const level = require('level')
 const sub = require('subleveldown')
 const AutoIndex = require('level-auto-index')
 const { Type } = require('@avro/types')
-const discovery = require('hyperdiscovery')
+const SwarmNetworker = require('corestore-swarm-networking')
 const crypto = require('hypercore-crypto')
 const DatEncoding = require('dat-encoding')
 const debug = require('debug')('p2pcommons')
@@ -105,11 +105,8 @@ class SDK {
     this.verbose = opts.verbose || false
     this.dbPath = opts.dbPath || this.baseDir
     // start hyperswarm
+    this.networkers = []
     this.disableSwarm = !!opts.disableSwarm
-    if (!this.disableSwarm) {
-      debug('constructor starting swarm')
-      this.swarm = discovery(Object.assign({}, DEFAULT_SWARM_OPTS, opts.swarm))
-    }
   }
 
   allowedProperties () {
@@ -123,6 +120,18 @@ class SDK {
     if (appType === 'profile') {
       return this.profileType
     }
+  }
+
+  _unflatten (data) {
+    if (typeof data.p2pcommons === 'object') return data
+    const { title, description, url, links, ...p2pcommons } = data
+    return { title, description, url, links, p2pcommons }
+  }
+
+  _flatten (data) {
+    if (typeof data.type === 'string') return data
+    const { p2pcommons, ...rest } = data
+    return { ...rest, ...p2pcommons }
   }
 
   async ready () {
@@ -234,26 +243,43 @@ class SDK {
     debug(`init storageLocation ${datOpts.datStorage.storageLocation}`)
 
     await ensureDir(datOpts.datStorage.storageLocation)
-    const archive = dat.create(datOpts.datStorage.storageLocation, publicKey, {
-      persist: this.persist,
-      storageFn: this.storage,
-      hyperdrive: {
-        secretKey
-      },
-      storageOpts: { ...datOpts.datStorage }
-    })
+    const { drive: archive, driveStorage } = dat.create(
+      datOpts.datStorage.storageLocation,
+      publicKey,
+      {
+        persist: this.persist,
+        storageFn: this.storage,
+        hyperdrive: {
+          secretKey
+        },
+        storageOpts: { ...datOpts.datStorage }
+      }
+    )
     await archive.ready()
 
     const hash = archive.key.toString('hex')
 
     if (!this.disableSwarm) {
-      this.swarm.add(archive)
+      const networker = new SwarmNetworker(driveStorage, {
+        announceLocalAddress: true
+      })
+      this.networkers.push(networker)
+      debug('swarm listening...')
+      networker.listen()
+      debug(`swarm seeding ${archive.discoveryKey.toString('hex')}`)
+      networker.seed(archive.discoveryKey)
 
+      dat.replicateFeed(driveStorage, archive.metadata)
       archive.once('close', () => {
-        debug(`closing archive ${archive.publicKey}...`)
         const discoveryKey = DatEncoding.encode(archive.discoveryKey)
-        this.swarm.leave(discoveryKey)
-        this.swarm._replicatingFeeds.delete(discoveryKey)
+        if (this.verbose) {
+          console.log(
+            `closing archive ${DatEncoding.decode(archive.discoveryKey)}...`
+          )
+        }
+        networker.unseed(discoveryKey)
+        // this.swarm.leave(discoveryKey)
+        // this.swarm._replicatingFeeds.delete(discoveryKey)
       })
     }
 
@@ -295,7 +321,8 @@ class SDK {
     if (this.verbose) {
       console.log(`Saved new ${datJSON.p2pcommons.type}, with key: ${hash}`)
     }
-    return datJSON
+    // Note(dk): flatten p2pcommons obj in order to have a more symmetrical API
+    return this._flatten(datJSON)
   }
 
   async saveItem ({ isWritable, lastModified, version, datJSON }) {
@@ -312,12 +339,14 @@ class SDK {
     const storageOpts = {
       storageLocation: join(this.baseDir, datJSON.url.toString('hex'))
     }
-    const archive = dat.open(
+    const { drive: archive, driveStorage } = dat.open(
       datJSON.url.toString('hex'),
       undefined,
       storageOpts
     )
     await archive.ready()
+    dat.replicateFeed(driveStorage, archive.metadata)
+
     let stat
     if (isWritable) {
       await writeFile(join(datJSONDir, 'dat.json'), JSON.stringify(datJSON))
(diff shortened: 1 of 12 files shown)


index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+      "'string' or Buffer",
+      destKey
+    )
+    // fetch source and dest
+    // 1 - try to get source from localdb
+    const source = await this.get(DatEncoding.encode(sourceKey))
+    if (!source) {
+      // 2 - if no module is found on localdb, then fetch from hyperdrive
+      // something like:
+      // const sourceDat = dat.open(source) // NOTE(dk): be sure to check sparse options so we only dwld dat.json
```
chartgerink: This one could also be a sparse clone onto the local machine, creating `~/.p2pcommons/<hash>+<version>/dat.json` only. This might streamline with other functionality and allow you to reuse code?

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+      // await sourceDat.ready()
+      // 3 - after fetching module we still need to read the dat.json file
+      // try {
+      //   source = await sourceDat.readFile('dat.json')
+      // } catch (err) {
+      //   throw new Error('Module not found')
+      // }
+      // 4 - populate localdb with the new module ??? VALIDATE FIRST
```
chartgerink: Do you mean "validate" as in whether the module is up to spec?

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
     return open(main)
   }
 
+  async register (sourceKey, destKey) {
```
chartgerink: `sourceKey` is of `type: content` and `destKey` of `type: profile`? I think we can specify those conditions.

The content is registered to the profile, in my thinking.

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+    }
+    const sourceType = this._getAvroType(source.type)
+    const sourceValid = sourceType.isValid(source)
+    if (!sourceValid) {
+      throw new Error('Invalid module')
+    }
+
+    // Note(dk): at this point is safe to save the new modules if necessary
+
+    if (dest.authors.length === 0) {
```
chartgerink: In my thinking, `dest` does not have the `authors` name/value because it is of `type: profile`. The `sourceKey` of `type: content` does have this.

`dest` has the `contents` name/value pair?

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+      throw new Error('Authors is empty')
+    }
+    // verify content first
+    const verified = await this.verify(source)
```
chartgerink: I look forward to seeing this in action :)

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+
+    if (!verified) {
+      throw new Error('source module does not met the requirements')
+    }
+
+    // register new content
+    dest.contents.push(source.url)
```
chartgerink: Here you use `dest.contents`, so I think on line 485 might be a bug :beetle:

index.js
```diff
@@ -415,6 +419,94 @@ class SDK {
...
+      typeof destKey === 'string' || Buffer.isBuffer(destKey),
+      ValidationError,
+      "'string' or Buffer",
+      destKey
+    )
+    // fetch source and dest
+    // 1 - try to get source from localdb
+    const source = await this.get(DatEncoding.encode(sourceKey))
+    if (!source) {
+      // 2 - if no module is found on localdb, then fetch from hyperdrive
```
chartgerink: I look forward to hearing how the version specific cloning goes. I remember messing around with this and this example worked but I'm not sure whether it allowed for version specific syncing (I don't know whether that's even possible)

https://github.com/mafintosh/mirror-folder/issues/16#issuecomment-397021772

index.js
dpaez: mmmh no, I  was just thinking to discuss if a fetched (from the hyperdrive) module should be saved in the local db? I guess we should save it.

index.js
dpaez: Yes, we should only do a sparse clone. I have to setup the sparse cloning options at this point and check the logic for creating such a directory `/<hash>+<version>/`

index.js
dpaez: > sourceKey is of type: content and destKey of type: profile? I think we can specify those conditions.
How should we specify this? Because since those are keys, we don't know anything about them before fetch (from the local db if they are there or from the hyperdrive). There is an assertion check after we fetch the content of each key for the type.

> The content is registered to the profile, in my thinking.
I guess we are on the same page here (I also see the profile as the target), I've just used the `source` and `dest` terms because I saw them on the [modules spec](https://github.com/p2pcommons/specs/blob/master/module.md#registration)

index.js
dpaez: > In my thinking, dest does not have the authors name/value because it is of type: profile. The sourceKey of type: content does have this.

you are right I'm going to fix this, `dest` does not have an `authors` field. In contrast, `content` type modules has this key. 

> dest has the contents name/value pair?

Yes.

So here we should check if `source.authors.length === 0`?

index.js
dpaez: :D

index.js
dpaez: I think by the line 485 you mean the `dest.authors.length` comparison? If that is the case, yes that is a bug. We should check `source.authors.`

index.js
dpaez: i'm thinking of doing an `hyperdrive.checkout(version)` here and then maybe mirroring (is it necessary?)

index.js
chartgerink: We could validate after fetch and just before registering?

index.js
chartgerink: Yes :)

index.js
dpaez: @chartgerink I just revisited this and I think we are a bit "restricted" to save a new "external" module, this is because two things:
1. `init` method currently does not accept an `url` or key as a param. An external module would already have a url.
2. `set` method can't modify the url. This is a restricted param. users can only update certain params. I guess I can add a `force` param to accomplish this but I'm open for suggestions.

index.js
chartgerink: Sorry, totally missed this comment. We can do checkout and then create a local copy under `~/.p2pcommons/<hash>+<version>`, for sure :)

If that is not what you meant, could you help me understand what you *did* mean?

lib/dat-helper.js
```diff
@@ -110,7 +110,12 @@ const importFiles = async (archive, src = './', opts = {}) => {
   })
 }
 
-const open = (key, location, storageOpts = {}) => {
+const open = (
+  key,
+  location,
+  storageOpts = {},
+  hyperdriveOpts = { sparse: true }
```
chartgerink: I'm getting ahead of things, simply want to mention that setting options is something in `v0.4.0` and ask whether you have ideas about how to do that? I guess we could pass a separate function, but it's relevant for the interoperability spec how we're storing this and where.

index.js
dpaez: yeah, I guess we are on the same page here :D

lib/dat-helper.js
dpaez: Yeah, I saw that setting options task was planned for `v0.4.0` but wasn't sure what specifically involves such a task. At the same I have to test/try the basic of registering and some sort of networking (basic swarming) would be required. We can play a bit with options like `sparse` here and improve/refactor how those options are handled for `0.4.0`.

chartgerink: Sounds perfect :hugs:

index.js
```diff
@@ -347,7 +380,7 @@ class SDK {
     })
   }
 
-  async filter (feature, criteria) {
+  async filter (feature, criteria, dbitem = false) {
```
juliangruber: boolean arguments maybe make more sense in TypeScript, but in JS `filter(feature, criteria, true)` isn't as descriptive as eg `filter({ feature, criteria, dbItem: true })`

index.js
```diff
@@ -362,44 +395,48 @@ class SDK {
...
+          out.push(dbitem ? v : v.rawJSON)
         }
       })
       s.on('end', () => resolve(out))
       s.on('error', reject)
     })
   }
 
-  async listContent () {
+  async listContent (dbitem = false) {
```
juliangruber: same as with `filter`

index.js
juliangruber: The `dbitem` option is there for when you need extra fields like `isWritable`, right?

juliangruber: maybe instead we can change the api to something like this, where the return type is always the same:

```js
async filter (feature, criteria) {
  // ...
  out.push({ dbItem, meta })
}
```

Basically this would decouple the db information from the rawJSON by making it two separate objects.

index.js
```diff
@@ -415,6 +452,97 @@ class SDK {
...
+      sourceKey
+    )
+    assert(
+      typeof destKey === 'string' || Buffer.isBuffer(destKey),
+      ValidationError,
+      "'string' or Buffer",
+      destKey
+    )
+    // fetch source and dest
+    // 1 - try to get source from localdb
```
juliangruber: this

1) get from local
2) get from hyperdrive

could be a good refactor into its own function

index.js
juliangruber: `source` and `dest` are a bit confusing to me, what about calling these parameters `contentKey` and `profileKey`?

index.js
dpaez: I like this improvement. It can be a breaking change for the cli though. Just for context, the `dbItem` param was added after the fact we realize that we need to keep that extra information, so sooner or later this little fix/refactor should happen. 😄

juliangruber: breaking changes are fine at this stage :)

dpaez: done in latest commits

index.js
dpaez: I'm going to move this refactors into their own issues to be done after 0.3 is out.

dpaez:
Alright, I think `register` is done for basic usage.  
Registering external modules is a WIP (working on networking and swarm creation using hyperswarm) . I think that feature will be added in upcoming 0.3.x versions. 
I don't want to hold this more time, so we can continue dogfooding 🐶 and getting more feedback.

index.js
dpaez: This becomes a separate issue for better tracking #40

PR #37, opened 2019-11-12
Update: add main param to init method

### LINKED ISSUE #36: unable to set .main in init method
Eyo!

This works for me:

```js
const { url } = await p2p.init({ type: 'content', title: 't', description: 'd' })
await p2p.set({ url, main: 'file' })
```

This doesn't (the main field won't be set):

```js
await p2p.init({ type: 'content', title: 't', description: 'd', main: 'file.txt' })
```

I think it would be nice if the `init` method supported all of the options too

dpaez:
this closes #36

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,26 +28,11 @@ const DEFAULT_SWARM_OPTS = {
   extensions: []
 }
 
-const createDatJSON = ({
-  type,
-  title,
-  subtype = '',
-  description = '',
-  main = '',
-  url = ''
-}) => {
-  assert(typeof type === 'string', ValidationError, 'string', type)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -42,7 +43,7 @@ const createDatJSON = ({
   obj.title = title
   obj.description = description
   obj.url = url
-  obj.main = ''
+  obj.main = main
```
juliangruber: instead of manually constructing `obj` from the parameters, what about using the parameters _as_ `obj` and then only augmenting it, eg to add the `license`?

dpaez: updated

Issue #41, opened 2019-11-20
unable to update profile module


juliangruber:
Given a `profile` module and sdk `v0.3.0`, this fails:

```js
await p2p.set({
  url: '7cc2f007a10bdd9c6355428642524c7cf19a331a5ed0c0408582ed478b141d69',
  title: 'name',
  description: 'desc'
})

// throws
// Invalid <RecordType>
```

juliangruber:
With a `content` module everything works just fine

dpaez:
Will look into it!

dpaez:
Hi @juliangruber I made some updates to the `set` test, now we are updating a profile module too, but tests are passing. Is this a similar scenario to yours?

juliangruber:
Yes thank you, the bug was on my side 🙏

PR #43, opened 2019-11-20
update tests: set profile module

dpaez:
Related to #41


juliangruber:
interesting! then the error must be somewhere else

juliangruber:
ok I figured it out, this was due to me inheriting from this module which broke some assumption. I'm now switching from inheritance to composition to fix this. thank you for writing this test :)

PR #54, opened 2019-12-13
Fix: register duplicates

### LINKED ISSUE #50: Registering content multiple times leads to duplicates
As a result of dogfooding [`@hypergraph-xyz/cli](https://github.com/hypergraph-xyz/cli/pull/56), I was mucking around with the registration functionality.

I re-registered content twice (total registration is three times) and got **nine** registered links. Maybe we should add a check to prevent repeat additions of equivalent Dat keys (both versioned and unversioned, regardless of #48)

```js
"contents": [
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d",
    "0ac8757f9adaf108ef9719225e2674adf12657640045f37ac4658f78f8ebd43d"
  ]
```

dpaez:
This PR closes #50

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -375,11 +375,11 @@ test('register - local contents', async t => {
   // update author on content module
   await p2p.set({ url: content1.url, authors })
 
-  await p2p.register(`${content1.url}+${metadata1.version}`, profile.url)
+  await p2p.register(`${content1.url}+${metadata1.version + 1}`, profile.url)
   const { rawJSON } = await p2p.get(profile.url)
   t.same(
     rawJSON.contents,
-    [content1.url],
+    [`${content1.url}+${metadata1.version + 1}`],
     'registration results in the addition of a dat key to the contents property of the target profile'
   )
   await p2p.destroy()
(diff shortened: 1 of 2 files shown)


tests/index.js
```diff
@@ -375,11 +375,11 @@ test('register - local contents', async t => {
   // update author on content module
   await p2p.set({ url: content1.url, authors })
 
-  await p2p.register(`${content1.url}+${metadata1.version}`, profile.url)
+  await p2p.register(`${content1.url}+${metadata1.version + 1}`, profile.url)
   const { rawJSON } = await p2p.get(profile.url)
   t.same(
     rawJSON.contents,
-    [content1.url],
+    [`${content1.url}+${metadata1.version + 1}`],
```
chartgerink: Is the +1 included because of writing the author to the `p2pcommons.authors` field? I wonder how this scales with multiple authors later, if so.

dpaez: yes, a couple of lines before that `+1` we did call `set` for _manually_ updating the authors field. We could do a new `get` to obtain the latest version but I was a bit lazy and since we are only doing one save I could figure it out the next version number.

chartgerink: Makes sense :) I'll make a note and we can revisit when it is relevant.

PR #67, opened 2020-01-28
Fix: drive management

### LINKED ISSUE #66: re-open old modules
Related to https://github.com/hypergraph-xyz/cli/issues/65

Current CLI PR shows an error when it tries to read a previously created (by a different instance) drive. 
The error indicates: `Error: This feed is not writable. Did you create it?` but the instance should be able to access the drive.

dpaez:
This PR attempts to fix #66 by removing any prior usage of `keyPair`s and relying only on namespaces for hyperdrive creation. 

It also adds re-seeding capabilities by storing previously created discovery keys in the local db.


juliangruber:
Nice work! What about adding a test using child processes to guard us against regressions?

dpaez:
Ohh that would be nice. The child process should consume the sdk and run
some commands? Something like that?

juliangruber:
Yeah I think so! One child could create a module, the other update it?

dpaez:
Yeah, on it, will ask for your review when its ready 👍

PR #78, opened 2020-02-06
New: unpublish method

### LINKED ISSUE #73: New method: unpublish content
**User story**
As a user of a p2pcommons implementation
I want to remove content from my profile
so that I can correct mistakes and keep my profile organised

**Acceptance criteria**
- The user can remove content from their own profile in a way that ensures that profile and content modules remain valid

**Possible implementations**
_Unpublish_ would be like the opposite action of current register method

So the API could be like `unpublish(contentKey, profileKey)` ? This means that the user already known the content and profile.

Another option is `unpublish(contentKey)`. This would remove the content from all the `contents` field of all the profiles.

dpaez:
This PR adds a new method: `unpublish`. It also renames the register method to `publish`. 

Closes #73 

Notes: 
- the method is checking if the profile is writable, if it isn't it will return (displays a message only if verbose mode is enabled). We can throw an error too.
- the SDK is not keeping a single "user profile". It creates modules that can be `content` or `profile`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ Remove content from a profile.
 
 > _async_ `delete(key: string or buffer)`
 
-Remove module from local db and seed db. If it was open in memory, its closed.
+Remove module from local db and seed db. If it was open in memory, its closed. Note: While this will stop the file from being seeded, that does not means that the content won't still be available on the network. This is due to the P2P file sharing dynamics.
 
 ### destroy
 
--- a/README.md
+++ b/README.md
@@ -139,11 +139,23 @@ Used to obtain a file descriptor from the `main` file of a module.
 - type: indicates the module type to retrieve. Allowed values: `profile`, `content`.
 - key: represents the module key (`url`) to be looked for. It is the buffer archive key `.toString('hex')`
 
-### register
+### publish
 
-> _async_ `register(contentKey: string or buffer, profileKey: string or buffer)`
+> _async_ `publish(contentKey: string or buffer, profileKey: string or buffer)`
 
 Register new content into a profile. The new content is added to the profile's `p2pcommons.contents`.
+
+### unpublish
+
+> _async_ `unpublish(contentKey: string or buffer, profileKey: string or buffer)`
+
+Remove content from a profile.
+
+### delete
+
+> _async_ `delete(key: string or buffer)`
+
+Remove module from local db and seed db. If it was open in memory, its closed.
 
 ### destroy
 
--- a/index.js
+++ b/index.js
@@ -108,26 +108,14 @@ class SDK {
     debug(`swarm enabled? ${!this.disableSwarm}`)
   }
 
-  _replicate (socket, info, handle) {
-    const isInitiator = !!info.client
-    const protocolStream = protocol({ live: true })
-    if (isInitiator) {
-      handle(protocolStream, info.peer.topic)
-    } else {
-      protocolStream.on('feed', discoveryKey => {
-        handle(protocolStream, discoveryKey)
-      })
-    }
-    pump(socket, protocolStream, socket, err => {
-      debug(err.message)
-      if (this.verbose) {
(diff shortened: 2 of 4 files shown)


juliangruber:
> * the method is checking if the profile is writable, if it isn't it will return (displays a message only if verbose mode is enabled). We can throw an error too.

It would be surprising to me if an error isn't thrown. Is there a reasoning why not to throw?

> * the SDK is not keeping a single "user profile". It creates modules that can be `content` or `profile`.

SGTM!

README.md
```diff
@@ -139,12 +139,24 @@ Used to obtain a file descriptor from the `main` file of a module.
...
+
+> _async_ `unpublish(contentKey: string or buffer, profileKey: string or buffer)`
+
+Remove content from a profile.
+
+### delete
+
+> _async_ `delete(key: string or buffer)`
+
+Remove module from local db and seed db. If it was open in memory, its closed.
```
juliangruber: Maybe add a note that this won't reliably remove it from the network?

index.js
```diff
@@ -420,12 +406,9 @@ class SDK {
...
 
-    if (this.verbose) {
-      // Note(dk): this kind of output can be part of the cli
-      console.log(
-        `Initialized new ${datJSON.p2pcommons.type}, dat://${publicKeyString}`
-      )
-    }
+    this._log(
+      `Initialized new ${datJSON.p2pcommons.type}, dat://${publicKeyString}`
+    )
```
juliangruber: nice refactors :)

index.js
```diff
@@ -865,9 +857,9 @@ class SDK {
    * @param {(String|Buffer)} contentKey - a dat url
    * @param {(String|Buffer)} profileKey - a dat url
    */
-  async register (contentKey, profileKey) {
```
juliangruber: the comment above still says register

index.js
```diff
@@ -950,7 +940,7 @@ class SDK {
       url: profile.url,
       contents: profile.p2pcommons.contents
     })
-    debug('register: profile updated successfully')
+    debug('publish: profile updated successfully')
```
juliangruber: why do we have `debug` and `this._log`? this line looks like something I would expect to see in verbose mode

index.js
```diff
@@ -982,6 +972,67 @@ class SDK {
     }, Promise.resolve(true))
   }
 
+  /**
+   * unpublish content from users profile
```
juliangruber:
```suggestion
   * unpublish content from a user's profile
```

index.js
```diff
@@ -982,6 +972,67 @@ class SDK {
...
+    const { module: profile, metadata } = await this._getModule(profileKey)
+
+    if (!metadata.isWritable) {
+      this._log('profile is not writable', 'warn')
+      return
+    }
+
+    if (!profile) {
+      this._log(`profile with key ${DatEncoding.encode(profileKey)} not found`)
+      return
```
juliangruber: this is also surprising, I expected an error here

README.md
dpaez: good call. will mention that. Maybe it will be interesting to add a link with the why... but yeah

index.js
dpaez: Oh yeah the SDK is becoming a bit large, so these refactors are necessary. Also at one point I'd like to improve the API using the API that you built for the CLI.

index.js
dpaez: yeah I expect to have more of these. I mean messages that can be a log and not a debug message. I see debug messages as something more internal, useful for developers maybe. But the line can be blurry.

dpaez: coming back to this line, yes, it's a good change for a log not a debug :)

index.js
dpaez: Oh I wasn't sure. But I'll use your comments as feedback and I will add the `throw`s. Thanks julian!

PR #103, opened 2020-03-02
Update: verify method

### LINKED ISSUE #97: verify method
**User story**
As an author
I want a distinction to be made between content that is verified to be mine and content that is not
so that I can have control over the work officially connected to my profile

**Acceptance criteria *(mostly from [@p2pcommons/specs/modue.md](https://github.com/p2pcommons/specs/blob/master/module.md))***
- The origin module to verify MUST be of `p2pcommons.type: content` and verification MUST occur on a specific version (i.e., versioned Dat archive key).
- The destination module in verification MUST be of `p2pcommons.type: profile` and SHOULD be the latest version (i.e., non-versioned Dat archive key). Verification MUST NOT be based on the outcome of verification across multiple versions (i.e., verification MUST be version specific).
- A version specific origin module MUST be deemed verified when all its `p2pcommons.authors` (destination) modules include the versioned origin module in their `p2pcommons.contents` property.
- The response when either or both of the modules are not available to be verified must be distinguishable from the response that occurs when a module is deemed not verified (for example by returning an error).

dpaez:
This PR closes #97 

It adds some improvements on the verify method, and updates the verify, publish and unpublish tests.
Note: `publish` does not call `verify`. From my understanding of the module spec, it's not necessary (of course if this assumption is wrong we can change it).

Nevertheless, there is a check on `publish` that validates the content module's authors field contains the profile involved.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1052,7 +1052,6 @@ class SDK {
       if (download) {
         dwldHandle = dat.downloadFiles(moduleVersion, folderPath)
       }
-      // store the module in localdb for future fetching
 
       let lastMeta
       try {
@@ -1060,6 +1059,7 @@ class SDK {
         lastMeta = metadata
       } catch (_) {}
 
+      // Note(dk): only update localdb if fetched module is more recent
       if (!lastMeta || stat.lastModified > lastMeta.lastModified) {
         await this.localdb.put(DatEncoding.encode(module.url), {
           isWritable: moduleVersion.writable,
--- a/index.js
+++ b/index.js
@@ -1212,9 +1212,8 @@ class SDK {
     if (module.p2pcommons.authors.length === 0) return false
     return module.p2pcommons.authors.reduce(async (prevProm, authorKey) => {
       const prev = await prevProm
-      // Note(dk): what if authorKey is not present on local db. fetch from swarm?
-      const { rawJSON: profile } = await this.get(authorKey)
-      return prev && profile.contents.includes(datUrl)
+      const { module: profile } = await this.clone(authorKey, null, false)
+      return prev && profile.p2pcommons.contents.includes(datUrl)
     }, Promise.resolve(true))
   }
 
--- a/index.js
+++ b/index.js
@@ -144,6 +144,15 @@ class SDK {
     )
   }
 
+  assertVersionedUrl (datUrl) {
+    assert(typeof datUrl === 'string', ValidationError, 'string', 'datUrl')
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -1158,29 +1189,32 @@ class SDK {
...
-    // TODO(dk): check versions
-    if (source.p2pcommons.authors.length === 0) return false
-    return source.p2pcommons.authors.reduce(async (prevProm, authorKey) => {
+    if (module.p2pcommons.authors.length === 0) return false
+    return module.p2pcommons.authors.reduce(async (prevProm, authorKey) => {
       const prev = await prevProm
       // Note(dk): what if authorKey is not present on local db. fetch from swarm?
       const { rawJSON: profile } = await this.get(authorKey)
-      return prev && profile.contents.includes(source.url)
+      return prev && profile.contents.includes(datUrl)
```
chartgerink: I'm not sure my code-reading is complete, but what does this do in case there are multiple authors?

It should loop over all, ultimately, and I'd like to check with you whether it does. :)

dpaez: Mmmh good call. Let me see if I can update test to reflect this.

dpaez:
@chartgerink I've just updated tests and added one that shows an scenario with multiple authors. What do you think?

dpaez:
This also closes #105

chartgerink:
Looking good! 🎉

jameslibscie:
We discussed verification during `publish` in the dev meeting this afternoon. It makes sense to us to leave it out for a few reasons:
- Verifying *before* publishing will result in `unverified` for first-time publishes, even though the publish should succeed
- Verifying *after* publishing doesn't prevent modules from becoming `unverified` unless you then turn back the change
- It can be left up to the implementation to decide whether to verify after publishing and how to handle this

What could make sense is:
- Checking whether the modules are writable
- Checking whether the modules would result in `verified` content. I.e., either both modules are writable OR one of them is not writable, but already refers to the other module

But neither of these seem necessary to me. Non-writable will result in an error anyway and the result can be checked afterwards and the user warned by the implementation.

Do you feel like this accurately represents the discussion @dpaez?

dpaez:
@jameslibscie Yes. That clarifies our conversation 💯

chartgerink:
Sounds good @jameslibscie! I think we can definitely do some things in the UX of Hypergraph to optimise verification of modules by giving "warnings", but it indeed makes sense to have those methods separated in the `sdk-js`.

PR #113, opened 2020-03-09
fix ignored leveldb open errors. closes #112

### LINKED ISSUE #112: leveldb open errors are ignored
https://github.com/p2pcommons/sdk-js/blob/48c093061aa688dba8dfdebdcf865758934b05cc/index.js#L381-L386

If an open error is encountered, the program will continue and succeed the `ready()` method, although it should have thrown.

juliangruber:
this works locally, let's see what CI says


dpaez:
this is amazing

Issue #93, opened 2020-02-25
Unfollow specific versions


jameslibscie:
*User story*
As a user of a p2pcommons implementation
I want to be able to unfollow specific versions of profiles that I've followed
so that I can predict the application's behaviour if I'm following multiple versions of the same profile

*Acceptance criteria*
- The `unfollow` method should only unfollow the specific versioned or non-versioned key that is provided
- The `unfollow` method should provide a warning if a key (or version of a key) is provided that is not currently followed

*Background*
I've noticed some strange behaviour in how unfollowing works:
- If I follow an unversioned profile, I can unfollow it by specifying a versioned profile
- If I follow a versioned profile, I can unfollow it by specifying an unversioned profile
- If I follow a versioned profile, I can unfollow it by specifying a different version of the profile
- Unfollowing a profile I don't follow at all doesn't throw an error

Given the fact that it's possible to follow multiple versions of the same profile (including the unversioned profile), I think the unfollow method should require the specific versioned or non-versioned key that is being followed.

This could be seen as a bug, but I don't think I spec'd #75 this clearly enough for it to be classified as such :)

PR #115, opened 2020-03-09
Update: specs 0.2.3

### LINKED ISSUE #99: Implement specs 0.2.3
`v0.2.3` of the [@p2pcommons/specs](https://github.com/p2pcommons/specs) contains the following updates:
- [ ] `p2pcommons.authors`, `parents`, `follows` and `contents` must be arrays of unique strings (i.e., they may contain multiple versions of the same archive, if versioning is allowed at all, but not duplicates of particular versions or duplicates of non-versioned keys)
- [ ] `p2pcommons.follows` must not self-reference (at any version)
- [ ] `p2pcommons.parents` must not self-reference at current or future version (i.e., content at `v14` may have itself at `v13` or `v2` as a parent, but not `v14` or `v15`

dpaez:
This PR closes #99 

One of the major things with this PR is that we refactor `set` method so other similar calls like `follow` and `publish` pass through it. Also added a `force` option, used to skip the validations and modify the metadata with the information that we are passing. This is required for cases where we delete stuff, `unfollow` and `unpublish`, in these cases methods should implement validation.


dpaez:
This also closes #93

PR #140, opened 2020-03-19
Update: delete force option

### LINKED ISSUE #90: Optionally delete local files belonging to content module
**User story**
As a user of a p2pcommons implementation
I want to delete the files associated with a content module
so that I don't end up with stray files that are hard to clean up

**Acceptance criteria**
- Deleting local files is optional
- Deleting local files moves the entire Dat archive folder to Trash / Recycle Bin / OS equivalent

dpaez:
This PR closes #90 

In summary, it adds a new option to `delete` method, called `force` which accepts a boolean value. If `true` it will move the module folder to the trash bin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- new force option to delete method #90
 
 ## [0.5.6] - 2020-03-18
 ### Security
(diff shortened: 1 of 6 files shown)


jameslibscie:
@dpaez I'm wondering whether `force` is descriptive enough as a parameter, since we might also use this terminology to mean something like "ignore validation errors". How about something like `deleteFiles`? What do you think @juliangruber as you'll be using this?

juliangruber:
I agree with @jameslibscie, force isn't describing what's actually happening. You would use it more in the context of "do it anyway". `deleteFiles` sounds good to me

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- new force option to delete method #90
```
dpaez:
```suggestion
- new deleteFiles option to delete method #90
```

PR #147, opened 2020-03-30
Update: trim title

### LINKED ISSUE #146: trim whitespaces in title
**Describe the bug**

~According to https://github.com/p2pcommons/specs/issues/28. Currently the SDK accepts a `" "` as a valid title.~

This has changed to only address the differences/updates on the specs.
 
**To Reproduce**
~Create a module with title = `"' '"`~

**Expected behavior**
~empty title should throw an error~

dpaez:
This closes #146


tests/index.js
```diff
@@ -122,6 +122,25 @@ test('init: creation should throw a ValidationError', async t => {
   }
 })
 
+test('init: should throw a ValidationError with empty title', async t => {
+  t.plan(1)
+  const p2p = createDb()
+  const metadata = {
+    type: 'content',
+    title: '" "'
```
juliangruber: I guess this title would be valid as `"` is considered a non-whitespace character. I propose we discuss this in the spec first.

dpaez: Currently the sdk is throwing when it finds an empty title like `'       '`. So I dig a bit more and found that something like `'" "'` could pass in and that's why I try to sanitize the title first and the trim it.

jameslibscie:
@dpaez The specs have been updating regarding this now.

dpaez:
Updated

schemas/validation.js
```diff
@@ -32,15 +38,22 @@ class Title extends LogicalType {
       throw new ValidationError('non empty string', val)
     }
 
-    if (!this._pattern.test(val)) {
+    const trimmed = sanitize(val).trim()
```
juliangruber, line 41: what's the purpose of using the string sanitizer module?

juliangruber: if the user wants to use `""` they are allowed to (according to the spec)

dpaez: yeah, I'm going to remove the sanitizer.

chartgerink:
After discussing with @dpaez and testing in Hypergraph (no errors), closing this PR as its outdated and out of spec.

Issue #122, opened 2020-03-11
Research Hyperdrive Daemon


jameslibscie:
This is exploratory work to determine what the best approach is for setting up a Daemon for p2pcommons. We can use this issue to share and discuss findings.

**Background**
@dpaez: From my point of view, there are two approaches, first one is to try to re-use hyperdrive-daemon (https://github.com/andrewosh/hyperdrive-daemon) using the client (https://github.com/andrewosh/hyperdrive-daemon-client), this could make us create a thin layer around it maybe or not, it depends on how can we adapt our business logic regarding how we treat hyperdrives into this thing. The other approach is just to try to code our own, it won't be that different from the hyperdrive-daemon and we can easily adapt our own rules to it. If we go with this approach it should be clear the why, and we need to think in some sort of compatibility with the hyperdrive-daemon.

PR #153, opened 2020-04-25
Update: hyperdrive-daemon integration

### LINKED ISSUE #125: Connect Daemon to SDK
*Waiting for #122* ❗️ 

**User story**
As a content creator
I want my profile and content to continuously synchronize to the network
so that my work is more available and more quickly updated

**Acceptance criteria**
- [ ] The SDK can connect to the Daemon (for what?)

dpaez:
This PR makes use of the `hyperdrive-daemon` and delegates to it the hyperdrive management and swarming logic.

To achieve this the PR introduces some changes, the most significant one is the change of the storage path. Previously all the modules could be easily found inside the `~/.p2pcommons` dir. Now we are only keeping there our internal DB. 
Since the `hyperdrive-daemon` is in charge of dealing with all the drives it makes sense to store the modules there. But, in order to keep everything in one place (daemon can be used by many apps), we are creating a special drive and mounting ([hyperdrive mount](https://github.com/mafintosh/hyperdrive#hyperdrive-mounting)) all our modules inside. 

Some SDK constructor options have slightly changed their meaning (like disableSwarm) but they are mentioned in the README and the idea was to minimize the need of changes on the consumers.

This is related to #122 and closes #125


README.md
```diff
@@ -57,10 +57,10 @@ Returns a new instance of the sdk.
 ```javascript
 {
   baseDir: // defaults to '~/.p2pcommons'
-  persist: // Indicates if the drives should be persisted - defaults to true
-  storage: // Accepts a storage function - defaults to undefined
-  disableSwarm: // defaults to undefined
-  swarm: // swarm options - defaults to undefined
+  rootKey: // archive key. Used when more than one sdk instance wants to share an hyperdrive custom folder
+  persist: // Indicates if the drives should be persisted or stored in-memory. Defaults to `true`
```
juliangruber, line 61: This is misleading, it could also be interpreted that `true` will make it so the drive will be persisted or stored in memory.

juliangruber: I think this was better before

README.md
```diff
@@ -57,10 +57,10 @@ Returns a new instance of the sdk.
...
+  storage: // Indicates a path to the hyperdrives folder. Defaults to the `hyperdrive-daemon` default storage.
+  disableSwarm: // defaults to undefined. If `false` will set the `hyperdrive-daemon` `noAnnounce` flag.
```
juliangruber, line 63: Whether `noAnnounce` will be set is an implementation detail, can we describe what this _means_ to the user?

index.js
```diff
@@ -96,24 +95,24 @@ class SDK {
...
 
-    // start hyperswarm
-    this.disableSwarm = !!finalOpts.disableSwarm
-    if (!this.disableSwarm) {
-      this.swarmFn =
-        finalOpts.swarm && typeof finalOpts.swarm === 'function'
-          ? finalOpts.swarm
-          : (...args) => new Swarm(...args)
-    }
+    // NOTE(dk): the meaning of disabling the swarm is going to change, possibly to be something like noAnnounce
```
juliangruber, line 114: what about doing that change right now, since we're already at it?

index.js
```diff
@@ -1832,14 +1817,16 @@ class SDK {
...
-        this._log(err.message, 'error')
-      }
-      debug('swarm successfully closed')
+        await d.close()
+      } catch (_) {}
+    }
+
+    if (swarm && this.dht) {
+      debug('destroying dht')
+      await this.dht.destroy()
```
juliangruber, line 1829: reviewing the actual implementation is over the top of my head, so I'm going to be focusing on tests and documentation

lib/dat-helper.js
```diff
@@ -1,57 +1,93 @@
...
+    }
+  }
+
+  try {
+    client = new HyperdriveClient(endpoint, finalOpts.token, {
+      storage: finalOpts.storage
+    })
+    await client.ready()
+  } catch (err) {
+    await HyperdriveDaemonManager.start({
```
juliangruber, line 48: could we just pass `finalOpts` directly?

lib/dat-helper.js
```diff
@@ -1,57 +1,93 @@
...
+      port: finalOpts.port,
+      bootstrap: finalOpts.bootstrap,
+      storage: finalOpts.storage,
+      memoryOnly: finalOpts.memoryOnly,
+      noAnnounce: finalOpts.noAnnounce
+    })
+  }
+
+  daemon = true
```
juliangruber, line 57: a boolean variable would be less surprising if it were called something like `daemonStarted` or `daemonRunning` in my view

lib/dat-helper.js
```diff
@@ -1,57 +1,93 @@
...
 
-// NOTE(dk): this can be used to get or create.
-const create = async (store, key = null, opts = {}) => {
-  assert(typeof opts === 'object', ValidationError, 'object', opts)
+const stopDaemon = async () => {
+  await HyperdriveDaemonManager.stop()
+}
 
-  const finalHyperdriveOpts = { ...DEFAULT_DRIVE_OPTS, ...opts.hyperdrive }
+const startClient = async (endpoint, token, opts = {}) => {
```
juliangruber, line 69: what if there already is a `client` from `startDaemon`?

tests/index.js
```diff
@@ -8,39 +8,68 @@ const once = require('events.once')
...
+const testDHT = require('./utils/dht')
 const level = require('level')
 const { encode } = require('dat-encoding')
 
-const testSwarmCreator = (store, opts) => testSwarm(store, opts)
+const daemonStorage = tempy.directory()
 
 const defaultOpts = () => ({
-  swarm: false,
+  disableSwarm: true,
```
juliangruber, line 18: the option was called `disableSwarm` before too, was this not doing that?

tests/index.js
```diff
@@ -346,8 +375,8 @@ test('set: should throw validation error with invalid main', async t => {
   t.end()
 })
 
-test('set: can update main', async t => {
-  const p2p = createDb()
+test.skip('set: can update main', async t => {
```
juliangruber, line 378: why is this test skipped?

tests/index.js
```diff
@@ -584,27 +611,27 @@ test('update: check version change', async t => {
...
+  const { rawJSON: rawJSON2, metadata: metadata2 } = await p2p.get(key)
 
   t.same(rawJSON2.description, description)
   t.ok(
     metadata2.version > metadata1.version,
     'latest version should be bigger than previous version after update'
   )
   t.ok(
-    metadata2.lastModified > metadata1.lastModified,
+    metadata2.lastModified.getTime() > metadata1.lastModified.getTime(),
```
juliangruber, line 626: you can compare date objects like this too:

```js
const a = new Date(0)
const b = new Date(1)
console.log(b > a)
```

tests/index.js
```diff
@@ -683,43 +710,35 @@ test('list modules', async t => {
 })
 
 test('multiple writes with persistance', async t => {
```
juliangruber, line 712:
```suggestion
test('multiple writes with persistence', async t => {
```

tests/index.js
```diff
@@ -765,18 +781,7 @@ test('publish - local contents', async t => {
...
-    persist: false,
-    swarmFn: testSwarmCreator
-  })
-  const p2p2 = createDb({
-    swarm: true,
-    verbose: true,
-    persist: false,
-    swarmFn: testSwarmCreator
-  })
+  const [p2p, p2p2] = await createDb(2)
```
juliangruber, line 784: above these are called `p2p1` and `p2p2`

tests/utils/dht.js
```diff
@@ -0,0 +1,19 @@
+const once = require('events.once')
+const dht = require('@hyperswarm/dht')
+
+const BOOTSTRAP_PORT = 3301
```
juliangruber, line 4: could you add a comment why a test dht is useful?

README.md
dpaez: I can rephrase it to something like: Boolean that indicates wether to save your modules into disk (true) or in-memory (false). Defaults to `true`

Or just change it to a `memory: true|false` option. To switch into in-memory storage.

index.js
dpaez: Oh, I did it. We are passing down to the daemon the noAnnounce flag.

index.js
dpaez: Yeah. That was my idea, just to have some of your thoughts and ideas. Also discuss how to release this and I also need to try it with the isntalable daemon app.

tests/index.js
dpaez: Yes, it was doing that. The thing is that before we were in control of the swarm, it was in our codebase and now we are delegating that to the daemon so that's why some options doesn't have like 1:1 alternative. I can remove this option or use something similar like the `noAnnounce` option.

tests/index.js
dpaez: Oh good catch, this shouldn't be skipped 😮

lib/dat-helper.js
dpaez: nice, yeah going to change it

dpaez:
Updated :)

lib/dat-helper.js
dpaez: done

README.md
juliangruber: if persistence is disabled, it's obvious that state will be stored in memory, as that's what all programs do. So I don't think it needs to be mentioned.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -24,15 +24,17 @@ A clear and concise description of what you expected to happen.
...
 **Smartphone (please complete the following information):**
- - Device: [e.g. iPhone6]
- - OS: [e.g. iOS8.1]
- - Browser [e.g. stock browser, safari]
- - Version [e.g. 22]
+
+- Device: [e.g. iPhone6]
+- OS: [e.g. iOS8.1]
+- Browser [e.g. stock browser, safari]
+- Version [e.g. 22]
```
juliangruber, line 37: I don't think these changes belong in this PR

README.md
```diff
@@ -54,13 +56,14 @@ const p2p = new P2PCommons()
...
 - `opts` can include the following:
+
 ```javascript
 {
   baseDir: // defaults to '~/.p2pcommons'
-  persist: // Indicates if the drives should be persisted - defaults to true
-  storage: // Accepts a storage function - defaults to undefined
-  disableSwarm: // defaults to undefined
-  swarm: // swarm options - defaults to undefined
+  rootKey: // archive key. Used when more than one sdk instance wants to share an hyperdrive custom folder
```
juliangruber, line 63: is it a root key or an archive key? :P

juliangruber: what is a "hyperdrive custom folder"?

README.md
```diff
@@ -54,13 +56,14 @@ const p2p = new P2PCommons()
...
+  memory: // A boolean that indicates if modules should be stored in-memory. Defaults to `false`
```
juliangruber, line 64:
```suggestion
  memory: // A boolean that indicates if modules should be stored in-memory only. Defaults to `false`
```

README.md
```diff
@@ -54,13 +56,14 @@ const p2p = new P2PCommons()
...
+  storage: // Indicates a path to the hyperdrives folder. Defaults to the `hyperdrive-daemon` default storage.
+  disableSwarm: // Used to avoid announcing our modules to the DHT, preventing them to be `seeded`. Defaults to `false`
```
juliangruber, line 66:
```suggestion
  disableSwarm: // Used to disable announcing our modules to the DHT, preventing them to be "seeded". Defaults to `false`
```

README.md
```diff
@@ -54,13 +56,14 @@ const p2p = new P2PCommons()
...
 ```javascript
 {
   baseDir: // defaults to '~/.p2pcommons'
-  persist: // Indicates if the drives should be persisted - defaults to true
-  storage: // Accepts a storage function - defaults to undefined
-  disableSwarm: // defaults to undefined
-  swarm: // swarm options - defaults to undefined
+  rootKey: // archive key. Used when more than one sdk instance wants to share an hyperdrive custom folder
+  memory: // A boolean that indicates if modules should be stored in-memory. Defaults to `false`
+  storage: // Indicates a path to the hyperdrives folder. Defaults to the `hyperdrive-daemon` default storage.
```
juliangruber, line 65:
```suggestion
  storage: // Path to the hyperdrives folder. Defaults to the `hyperdrive-daemon` default storage location.
```

README.md
```diff
@@ -83,6 +86,7 @@ Retrieves raw datJSON item and metadata from the local db.
 - hash: represents the key (`url`) to be looked for. It is the buffer archive key `.toString('hex')`
 
 Returns an object with:
+
```
juliangruber, line 89: I love that except for the constructor the API staid the same <3

index.js
juliangruber: shouldn't this be addressed in this PR then?

package.json
```diff
@@ -4,12 +4,16 @@
...
     "format": "prettier-standard --format",
     "version": "chan release ${npm_package_version} --ghrelease && git add CHANGELOG.md",
     "prepublishOnly": "npm test"
   },
-  "precommit": "prettier-standard --lint --staged",
+  "husky": {
+    "hooks": {
+      "pre-commit": "prettier-standard --lint --staged"
+    }
+  },
```
juliangruber, line 16: I believe this is also outside the scope of this PR

schemas/profile.json
```diff
@@ -69,7 +69,7 @@
               "symbols": ["content", "profile"]
             }
           },
-          {"name": "subtype", "type": "string"},
+          { "name": "subtype", "type": "string" },
```
juliangruber, line 72: it's more difficult to review with these linting changes :P

tests/utils/dht.js
```diff
@@ -0,0 +1,27 @@
...
+const BOOTSTRAP_URL = `localhost:${BOOTSTRAP_PORT}`
+
+/*
+ * In memory DHT
+ * Goal: speed up tests by avoiding announce everything to the real DHT
```
juliangruber, line 9:
```suggestion
 * Goal: speed up tests by avoiding to announce everything to the real DHT
```

tests/utils/dht.js
```diff
@@ -0,0 +1,27 @@
...
+ * and reducing this way network usage.
```
juliangruber, line 10:
```suggestion
 * and reducing network usage this way.
```

dpaez:
I made some changes recently and simplified a bit the interaction with the daemon. I also used the `hypergraph/cli` to see how the integration works. I have to do some adjustments on the cli and it almost works, some tests like `follow/unfollow` `publish/unpublish` run partially and then they get blocked. 

Last week julian mentioned something that may work and it is to use fuse in combination with the new `import/export` daemon mechanism.  So I've been thinking how could this work on the SDK. This would change again our modules directory (currently living inside `~/.hyperdrive/storage/<hash>`).
So, if we add `fuse` support we can create a subdirectory called `p2pcommons`, this will live in `~/Hyperdrive/p2pcommons` and the user can work with their modules in this directory. In the other case, if the user is on windows or does not have fuse installed then we will switch to the import/export strategy. Here the working directory could look like this (we can re-use the .p2pcommons): `~/.p2pcommons/storage/<hash>`.
If you are OK with these changes, I can go ahead and try using this new setup. 👍

juliangruber:
> So, if we add `fuse` support we can create a subdirectory called `p2pcommons`, this will live in `~/Hyperdrive/p2pcommons` and the user can work with their modules in this directory.

Why this? I'm not sure people will even use that directory directly, since paths like `~/Hyperdrive/aababcacbab3434535` aren't user friendly. I think they will use the Hypergraph Desktop / CLI instead.

Also, with fuse, we control the mountpoints, so we can mount at `~/Hyperdrive/<module title>`.

dpaez:
@juliangruber yeah, actually using fuse and the import/export we could use better directory names. 

I mentioned the `~/Hyperdrive/p2pcommons/` directory because that's what I understood from the daemon docs. But, looking at the mount and create commands again it seems like we can do that, like mount something external into the shared `~/Hyperdrive` directory. 

Following that line, maybe we can assume a directory to be the default Hypergraph modules directory (and let the user specify/change it if they want to).

juliangruber:
> Following that line, maybe we can assume a directory to be the default Hypergraph modules directory (and let the user specify/change it if they want to).

If the daemon is already running, and eg not configured to use `~/Hyperdrive`, we won't be able to place the modules anywhere else using fuse. So we won't have full control over this anyways, right?

chartgerink:
Superceded by #160, closing this after discussion with @dpaez.

PR #158, opened 2020-05-21
Add html extension to spec link

chartgerink:
We did not include the html extension to the spec for the metadata, which would never resolve as of https://github.com/p2pcommons/specs/commit/b4cf9ca59f7d65eccd979e933fad8f1fa3b16d8e :)

Opening a PR for due process! I don't know whether we need to edit the schemas elsewehere either.


chartgerink:
Oh wait it does resolve so this PR can be regarded as unnecessary. 
https://p2pcommons.com/specs/module/1.0.0

Will close! 👍

Issue #165, opened 2020-06-25
update sodium-native


juliangruber:
Currently Hypergraph Desktop can't update its electron version because sodium-native@1 doesn't have prebuilds for the newer chrome versions.

Also, deleting the package-lock and reinstalling deps currently breaks the app, maybe because then different versions of sodium-native end up in the tree.

Thanks :)

dpaez:
closed with #166

PR #166, opened 2020-06-25
Update: hypercore-crypto

dpaez:
Related to #165

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - renamed (un)publish to (de)register
 - renamed dat.json to index.json
-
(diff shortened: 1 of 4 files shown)


dpaez:
> Has this package-lock been manually changed or is it `rm package-lock.json && npm install`?

Was `rm package-lock.json && npm install`

Now, I've updated the hypercore-crypto dep, to version 2.

juliangruber:
I think this still contains the pre-v3 sodium native, if you look at the package-lock. `npm ls sodium-native` should tell what brings it in (if I'm not mistaken)

dpaez:
yeah, `universal-dat-storage` includes this dep :( I can see if I can remove it now. FWIW this dep is no longer used in the latest version of the SDK (hyperdrive-daemon based)

dpaez:
Cool, I've managed to remove the universal-dat-storage dep, so now we are only using sodium-native@3

Issue #155, opened 2020-05-11
improve test coverage


juliangruber:
Hey Diego 👋,

As discussed in our office hour, we would like to increase the test coverage on the sdk, as often issues are found in the CLI when updating to a new sdk version.

We mentioned a couple potential strategies:

- a certain test coverage percentage
- regression tests
- test that the sdk is spec compliant
- integration tests

jameslibscie:
See also #154 for bringing the SDK up to spec and creating the validations that can then be fully tested.

jameslibscie:
Closed via #167

PR #167, opened 2020-06-26
Metadata validations + tests

### LINKED ISSUE #154: Complete metadata validation + atomic metadata validation
**User stories**
As a developer of a p2pcommons implementation
I want to be able to validate the current metadata (`dat.json`)
so that I can protect users against errors, even those created outside my application
and
I want to be able to validate user input (possible future metadata or snippets of metadata)
so that I don't have to implement validation methods in my application that may become out of sync with the SDK's validations

**Acceptance criteria**

- [x] There is a validation method that fully validates a set of metadata that is passed to it
  - [x] Context-specific validations are possible (for example, validation on `publish` should validate that the `main` file exists)
- [x] Each part of the metadata can be validated separately, to ensure that implementations can "test" specific changes before effecting them
- [x] Make sure tests check for specific errors

**Suggested implementation**

I discussed this with @juliangruber and it seems most user/dev-friendly to us to implement all the validation methods with the same two params, namely an object containing module metadata (`dat.json`) and an object containing additional metadata like `version`, `lastModified`, `isWritable`. This allows for context-specific validation where necessary.

For a simple validation, like `title`, where the validation is not context-dependent, passing `{ title: "Nice title" }` as metadata is enough and the param for additional metadata is ignored, but passing a full (current or "hypothetical") object with metadata also works. Each method only validates what is necessary to determine whether that part of the validation is successful. Altogether, all the partial methods make up a full validation of the module metadata.

**Suggested list of validations + error messages**

This list should be up to spec with module spec `0.3.2`

*Note: it might be valuable to add error codes to each validation as well. What do you think @juliangruber? I've quickly added suggestions*

`title`

- [x] Required. "Title is required", `title_required`
- [x] Type string. "Title must be a string", `title_type`
- [x] Between 1 and 300 characters long. "Title must be between 1 and 300 characters long", `title_length`
- [x] Does not consist of only whitespace. "Title may not consist of only whitespace", `title_whitespace`

`description`

- [x] Required. "Description is required", `description_required`
- [x] Type string. "Description must be a string", `description_type`

`url`

- [x] Required. "URL is required", `url_required`
- [x] Type string. "URL must be a string", `url_type`
- [x] Starts with `hyper://`. "URL must start with hyper:// protocol", `url_protocol`
- [x] Follows regex [`^(hyper:\/\/)(\w{64})$`](https://regex101.com/r/naEFVg/5). "URL must contain non-versioned Hyperdrive key", `url_invalidkey`
- [x] The key in the URL is equal to the module's Hyperdrive key. "URL must refer to the module's own Hyperdrive key", `url_incorrectkey`

`links`

- [x] Required. "Links object is required", `links_required`
- [x] Type object. "Links must be an object", `links_type`
- [x] All keys must be strings. "Links must be an object with string keys", `links_stringkeys`
- [x] All values must be arrays. "Links must be an object with array values", `links_arrayvalues`

`links.license`

- [x] Required. "License link is required", `links_license_required`
- [x] Type array. "License must be an array", `links_license_type`
- [x] Must contain one object. "License must contain one object", `links_license_type`
  - [x] This object must contain an `href` key. "License object must have an href key", `links_license_href`
  - [x] The `href` value must be equal to `https://creativecommons.org/publicdomain/zero/1.0/legalcode`. "License link must be equal to 'https://creativecommons.org/publicdomain/zero/1.0/legalcode'", `links_license_value`

`links.spec`

- [x] Required. "Spec link is required", `links_spec_required`
- [x] Type array. "Spec must be an array", `links_spec_type`
- [x] Must contain one object. "Spec must contain one object", `links_spec_type`
  - [x] This object must contain an `href` key. "Spec object must have an href key", `links_spec_href`
  - [x] The `href` value must be a valid URL. "Spec href must refer to a valid URL", `links_spec_validurl`

`p2pcommons`

- [x] Required. "p2pcommons object is required", `p2pcommons_required`
- [x] Type object. "p2pcommons must be an object", `p2pcommons_type`

`p2pcommons.type`

- [x] Required. "p2pcommons type is required", `p2pcommons_type_required`
- [x] Type string. "p2pcommons type must be an array", `p2pcommons_type_type`
- [x] Must be equal to `profile` or `content`. "p2pcommons type must be equal to 'profile' or 'content'", `p2pcommons_type_value`

`p2pcommons.subtype`

- [x] Required. "p2pcommons subtype is required", `p2pcommons_subtype_required`
- [x] Type string. "p2pcommons subtype must be a string", `p2pcommons_subtype_type`
- [x] Follows regex [`^\w*$`](https://regex101.com/r/hDRGfc/2). "p2pcommons subtype may only include standard alphanumeric characters", `p2pcommons_subtype_alphanumeric`

`p2pcommons.main`

- [x] Required. "p2pcommons main is required", `p2pcommons_main_required`
- [x] Type string. "p2pcommons main must be a string", `p2pcommons_main_type`
- [x] Must not refer to a `.dotfile`. "p2pcommons main may not be a .dotfile", `p2pcommons_main_dotfile`
- [x] Must be a relative path to a file within the Hyperdrive OR empty (regex [`^((?!\/)|(\.\/))(?!~|\.).*(?<!\/)$`](https://regex101.com/r/MZXJnK/1)). "p2pcommons main must be a relative path to a file within the Hyperdrive OR empty", `p2pcommons_main_relativepath` *Note: this regex also excludes some `.dotfiles`, but not all, so apply after the dotfile check*
- [x] Upon publishing, must refer to an existing file. "p2pcommons main file does not exist", `p2pcommons_main_filemissing`

*Note: Perhaps this validation should be split into 2 methods, something like `validateMainFormat()` and `validateMainExists()` or `validateMain()` and `validateMainFile()`*

`p2pcommons.avatar`

- [x] May only exist if content type is `profile`. "p2pcommons avatar may only exist for profiles", `p2pcommons_avatar_moduletype`
- [x] Type string. "p2pcommons avatar must be a string", `p2pcommons_avatar_type`
- [x] Must be a relative path to a file within the Hyperdrive OR empty (regex [`^((?!\/)|(\.\/))(?!~|\.).*(?<!\/)$`](https://regex101.com/r/MZXJnK/1)). "p2pcommons avatar must be a relative path to a file within the Hyperdrive OR empty", `p2pcommons_avatar_relativepath` *Note: this regex also excludes some `.dotfiles`, which are allowed for the avatar*

`p2pcommons.authors`

- [x] Required if content type is `content`. "p2pcommons authors is required for content", `p2pcommons_authors_required`
- [x] May only exist if content type is `content`. "p2pcommons authors may only exist for content", `p2pcommons_authors_moduletype`
- [x] Type array. "p2pcommons authors must be an array", `p2pcommons_authors_type`
- [x] Each item must be a string. "p2pcommons authors must be an array of strings", `p2pcommons_authors_type`
- [x] Must contain unique values. "p2pcommons authors must be unique", `p2pcommons_authors_unique`
- [x] Each item must be a non-versioned Hyperdrive key (regex [`^(\w{64})$`](https://regex101.com/r/GQOim5/2)). "p2pcommons authors may only contain non-versioned Hyperdrive keys", `p2pcommons_authors_invalidkey`

`p2pcommons.parents`

- [x] Required if content type is `content`. "p2pcommons parents is required for content", `p2pcommons_parents_required`
- [x] May only exist if content type is `content`. "p2pcommons parents may only exist for content", `p2pcommons_parents_moduletype`
- [x] Type array. "p2pcommons parents must be an array", `p2pcommons_parents_type`
- [x] Each item must be a string. "p2pcommons parents must be an array of strings", `p2pcommons_parents_type`
- [x] Must contain unique values. "p2pcommons parents must be unique", `p2pcommons_parents_unique`
- [x] Each item must be a versioned Hyperdrive key (regex [`^(\w{64})(\+\d+)$`](https://regex101.com/r/GQOim5/3)). "p2pcommons parents may only contain versioned Hyperdrive keys", `p2pcommons_parents_invalidkey`
- [x] If an item refers to the module's own Hyperdrive key, it must refer to a preceding version. "p2pcommons parents may not refer to current or future versions of itself", `p2pcommons_parents_selfreference` *Note: it IS allowed to refer to a current version -before- updating the version number, as when the update is performed, the proposed parent and the current version number will no longer align. After the new version is created, version > parent version*
- [x] Upon setting `parents` (but not at other times), it must refer to a content module that is published by at least one of its listed authors. "p2pcommons parents must be published", `p2pcommons_parents_notpublished`

`p2pcommons.follows`

- [x] Required if content type is `profile`. "p2pcommons follows is required for profiles", `p2pcommons_follows_required`
- [x] May only exist if content type is `profile`. "p2pcommons follows may only exist for profiles", `p2pcommons_follows_moduletype`
- [x] Type array. "p2pcommons follows must be an array", `p2pcommons_follows_type`
- [x] Each item must be a string. "p2pcommons follows must be an array of strings", `p2pcommons_follows_type`
- [x] Must contain unique values. "p2pcommons follows must be unique", `p2pcommons_follows_unique`
- [x] Each item must be a versioned OR non-versioned Hyperdrive key (regex [`^(\w{64})(\+\d+)?$`](https://regex101.com/r/GQOim5/4)). "p2pcommons follows may only contain Hyperdrive keys (versioned or non-versioned)", `p2pcommons_follows_invalidkey`
- [x] An item may not refer to the module's own Hyperdrive key at any version. "p2pcommons follows may not contain the profile's own Hyperdrive key", `p2pcommons_follows_selfreference`

`p2pcommons.contents`

- [x] Required if content type is `profile`. "p2pcommons contents is required for profiles", `p2pcommons_contents_required`
- [x] May only exist if content type is `profile`. "p2pcommons contents may only exist for profiles", `p2pcommons_contents_moduletype`
- [x] Type array. "p2pcommons contents must be an array", `p2pcommons_contents_type`
- [x] Each item must be a string. "p2pcommons contents must be an array of strings", `p2pcommons_contents_type`
- [x] Must contain unique values. "p2pcommons contents must be unique", `p2pcommons_contents_unique`
- [x] Each item must be a versioned OR non-versioned Hyperdrive key (regex [`^(\w{64})(\+\d+)?$`](https://regex101.com/r/GQOim5/4)). "p2pcommons contents may only contain Hyperdrive keys (versioned or non-versioned)", `p2pcommons_contents_invalidkey`

jameslibscie:
Contains:
- Full and partial exported metadata validations
- Tests for all new validations
- Replaced old validations where possible
- Updated `ValidationError` and added `TypeError` to replace old `ValidationErrors` where they remain
- Remove `hyper://` protocol for `p2pcommons` keys
- Updated readme 

Todo:
- [x] Complete validations + tests
- [x] Propose specs changes based on findings
- [x] Replace old validations with new validations
- [x] Migrate old remaining `ValidationErrors`
- [x] Merge `validateBeforeInit()` and `validateDraft()` into `validatePartial()`
- [x] Async functions
- [x] Export new validations
- [x] Make sure new tests are included
- [x] Fix `validateMainExists()` to work with versioned folder names
- [x] Remove `hyper://` protocol from `p2pcommons` keys + validations
- [x] Update `validateOnRegister()` + tests to include `validate()`
- [x] Try out with CLI or Desktop
- [x] Document validations in code & readme
- [x] Changelog
- [x] Object parameters
- [x] Clarify partial validations (present vs `undefined` or `null`)

Closes #154, #155 and #179


juliangruber:
Closes #186

README.md
```diff
@@ -187,13 +209,79 @@ If `deleteFiles` option is true, then the target folder will be moved to the tra
...
+Validates the p2pcommons object structure (not its contents)
+> _async_ `validateType(indexMetadata: object, unflatten: boolean)`
+> _async_ `validateSubtype(indexMetadata: object, unflatten: boolean)`
+> _async_ `validateMain(indexMetadata: object, key: string, p2pcommonsDir: string, unflatten: boolean)`
+Also checks the existence of the specified main file
+> _async_ `validateAvatar(indexMetadata: object, unflatten: boolean)`
+> _async_ `validateAuthors(indexMetadata: object, unflatten: boolean)`
+> _async_ `validateParents(indexMetadata: object, dbMetadata: object, key: string, unflatten: boolean)`
+> _async_ `validateFollows(indexMetadata: object, key: string, unflatten: boolean)`
+> _async_ `validateContents(indexMetadata: object, unflatten: boolean)`
```
juliangruber, line 250: I think this is not rendered as you would like it to be

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+ * @public
+ * @async
+ * @throws ValidationError
+ * @param {object} indexMetadata - metadata from index.json (=rawJSON)
+ * @param {object} dbMetadata - metadata from the database (=metadata)
+ * @param {string} key - versioned or unversioned Hyperdrive key
+ * @param {string} p2pcommonsDir - path to p2pcommons directory
+ * @param {[Boolean]} unflatten - set false to validate non-flattened metadata (i.e. directly read from index.json)
+ */
+const validate = (exports.validate = async (indexMetadata, dbMetadata, key, p2pcommonsDir, unflatten = true) => {
```
juliangruber, line 21: this method has enough arguments to switch to an object instead. Otherwise this order is difficult to get right.

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+/**
+ * Fully validates a module against the p2pcommons module specs
+ *
+ * @public
+ * @async
+ * @throws ValidationError
+ * @param {object} indexMetadata - metadata from index.json (=rawJSON)
+ * @param {object} dbMetadata - metadata from the database (=metadata)
+ * @param {string} key - versioned or unversioned Hyperdrive key
+ * @param {string} p2pcommonsDir - path to p2pcommons directory
```
juliangruber, line 18: could this have a default value?

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+ * @param {[Boolean]} unflatten - set false to validate non-flattened metadata (i.e. directly read from index.json)
```
juliangruber, line 19: what about changing this to `flat` which defaults to `false`?

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+  validateParents(indexMetadata, dbMetadata, key, unflatten)
+  validateFollows(indexMetadata, key, unflatten)
+  validateContents(indexMetadata, unflatten)
+  await validateMain(indexMetadata, key, p2pcommonsDir, unflatten)
+})
+
+/**
+ * Validates all present data against the p2pcommons specs.
+ * Can be used for validating unfinished modules.
+ * Empty p2pcommons.main will not be validated
```
juliangruber, line 43:
```suggestion
```

It already says above that empty fields won't be validated

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+ * @param {[string]} dbMetadata - metadata from the database (=metadata)
+ * @param {[string]} key - versioned or unversioned Hyperdrive key
+ * @param {[string]} p2pcommonsDir - path to p2pcommons directory
+ * @param {[Boolean]} unflatten - set false to validate non-flattened metadata (i.e. directly read from index.json)
+ */
+exports.validatePartial = async (indexMetadata, dbMetadata = { version: 0 }, key = '', p2pcommonsDir = '', unflatten = true) => {
+  if (unflatten) {
+    indexMetadata = _unflattenIndexMetadata(indexMetadata)
+  }
+  if (indexMetadata.title !== undefined) {
```
juliangruber, line 58: What if `.title === null`? I think this could either check for both, or do a `typeof` check

README.md
jameslibscie: Good catch, forgot to check this

lib/validate.js
jameslibscie: Is this something you feel strongly about? I'm willing to change it, but it's a lot of work. I considered this, but figured that you'd need hinting for all these methods anyway as they don't all require the same arguments. Lmk :)

lib/validate.js
jameslibscie: I didn't think it could, but I've separated the `p2pcommons.baseDir` logic into `lib/base-dir.js` so the main SDK and the validations both call that for deriving the default path. Of course a custom path is still possible, for the main SDK and for the validations.

lib/validate.js
jameslibscie: I've changed it to `flat`, but it still defaults to `true` as that is the format that the rest of the SDK works with. Do you think this should default to `false` (so it works directly with `index.json` content but requires a flag for SDK output) and if so, why?

lib/validate.js
jameslibscie: Clarified a little.

> Validates all present data against the p2pcommons specs. Can be used for validating unfinished modules. If p2pcommons.main is present, but empty, it will not be validated.

lib/validate.js
jameslibscie: Any specific reason for also skipping `null` values? And would you skip empty values too? This would help to work out the `main` exception to the rule, but could mean that some invalid fields (like an empty `title`) are mistakenly skipped?

lib/validate.js
juliangruber: I do feel strongly about this, JS not being a typed language long argument lists are really difficult to get right, and can easily introduce bugs.

lib/validate.js
juliangruber: Aah I now understand better..When would `flat: false` be useful? Is this used in the SDK internally?

lib/validate.js
juliangruber: Ah! Could we clarify what "present" data means? One could argue both ways whether `{ main: undefined }` is present or not. What about validating all keys that exist (as in `if (key in obj)`), or are not `typeof value === undefined`?

lib/validate.js
juliangruber: This is related to the comment above..in JS, `null` and `undefined` are used interchangeably for denoting a non existing value. For example, I would set `p2p.set({ key, main: null })` to try to remove the main file.

lib/validate.js
```diff
@@ -0,0 +1,1188 @@
...
+    'contents_type',
+    'p2pcommons.contents'
+  )
+}
+
+const _validateContentsUnique = (indexMetadata) => {
+  const uniqueValues = [...new Set(indexMetadata.p2pcommons.contents)]
+  assert(
+    indexMetadata.p2pcommons.contents === undefined ||
+        uniqueValues.length === indexMetadata.p2pcommons.contents.length,
```
juliangruber, line 1067: nice implementation!

lib/validate.js
jameslibscie: The SDK only uses flat metadata, but `index.json` is not flat. You could potentially end up with the situation that `index.json` is stored flattened (or edited by a user to become somewhat flattened) which would technically make it invalid. This parameter allows you to directly read `index.json` and validate the unflattened metadata. The only internal use is for tests, since I mostly use unflattened metadata for those.

juliangruber: If this is only used for tests I would not expose it, or at least not document it, and update the comment that this is an internal option. Maybe even prefix it with `_`

lib/validate.js
jameslibscie: Cool - done!

lib/validate.js
jameslibscie: Changed internal documentation, removed from readme and prefixed with _

lib/validate.js
jameslibscie: Clarified by specifying not null or undefined

lib/validate.js
jameslibscie: Null is now also skipped

lib/validate.js
jameslibscie: Thanks! Long live stackexchange 🤘🏻

lib/validate.js
juliangruber: Thanks James :)

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The SDK for the p2p commons",
   "main": "index.js",
   "scripts": {
-    "test": "node tests/index.js | tap-spec",
+    "test": "node tests/tap.js tests/index.js tests/validate.js | tap-spec",
```
juliangruber, line 7:
```suggestion
    "test": "tap tests/index.js tests/validate.js | tap-spec",
```

Not sure it's worth it introducing a custom test runner for a small test base

schemas/validation.js
```diff
@@ -70,70 +70,128 @@ class Path extends LogicalType {
...
+}
+
+class HyperKeyVersioned extends LogicalType {
+  constructor (attrs, opts) {
+    super(attrs, opts)
+    let pattern = /([a-f0-9]{64})(\+\d+)?$/
+    if (attrs.strict) {
+      pattern = /^([a-f0-9]{64})(\+\d+)$/
+    }
+    this._pattern = new RegExp(pattern)
```
juliangruber, line 176: `pattern` already is a RegExp, no need to instantiate it again

jameslibscie: Copy/paste of existing code - fixed here and in the code I copied from

package.json
jameslibscie: Fixed!

jameslibscie:
@dpaez Can you still review before merging? :)

PR #169, opened 2020-07-01
Update dat:// to hyper://

chartgerink:
This PR updates all mentions of the Dat protocol to the Hypercore protocol, in accordance with the new name [as announced recently](https://blog.datproject.org/2020/05/15/dat-protocol-renamed-hypercore-protocol/).


dpaez:
test are failing, possibly due to something in the schemas, I'm going to check that now

dpaez:
schema updated, but we are hitting an error with `dat-encoding` which is not expecting hyper:// urls. I can try and make a PR there but also would like to evaluate if we can replace or remove it.

jameslibscie:
Would it make sense to include #179 in this PR?

chartgerink:
I would suggest to do it in a separate PR because the tests are already failing here. We can't really test whether the fix to #179 would then pass or not.

jameslibscie:
> I would suggest to do it in a separate PR because the tests are already failing here. We can't really test whether the fix to #179 would then pass or not.

The reason I'm suggesting it is because it affects the same code. Having two parallel PRs would generate a bunch of conflicts. If we don't do them together in 1 PR, then I think we'd be best off doing them sequentially.

dpaez:
> schema updated, but we are hitting an error with `dat-encoding` which is not expecting hyper:// urls. I can try and make a PR there but also would like to evaluate if we can replace or remove it.

Luckily I was wrong with this, and there is no need to fork `dat-encoding`, shouldn't be a blocker. There were some `dat` remains here and there that were causing some errors. Those are fixed now. 👍

PR #189, opened 2020-07-14
release 0.6.3

dpaez:
New release :tada:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.6.3] - 2020-07-14
 ### Changed
-- rename dat:// with hyper:// #169
+- rename dat:// to hyper:// #169
 - pretty print JSON write-out #183 
 - SDK now emits EBUSY error as a warn event #185
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [0.6.3] - 2020-07-14
 ### Changed
 - rename dat:// with hyper:// #169
 - pretty print JSON write-out #183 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [0.6.3] - 2020-07-14
 ### Changed
 - rename dat:// with hyper:// #169
```
juliangruber, line 11:
```suggestion
- rename dat:// to hyper:// #169
```
```suggestion
- replace dat:// with hyper:// #169
```

juliangruber: Below it says "Use module property for spec #182". What does that mean?

README.md
```diff
@@ -224,7 +224,14 @@ Error object contains some useful properties:
 
 ## Release
 
-`npm publish`
+1. create a new branch like `{username}/release-{nextVersion}`
+2. `npm version {nextVersion}`
+    - this will prompt you with a new gh release window, you can mark it as _draft_ until PR is accepted.
+3. _acceptance_ & _merge_
```
juliangruber, line 230: does this mean to also create a pull request, push the branch and the tag?

dpaez: I'm not sure. I think npm version creates the tag, but we are publishing from `main`.

This is also a tentative process, open for improvement.

CHANGELOG.md
dpaez: > Below it says "Use module property for spec #182". What does that mean?

We added a new module called `spec.js`, it should help tracking the different versions of the spec that we are using. The problem was that we were using the wrong property for one of the specs.

README.md
juliangruber: Unless you use a regular merge, I think the tag will point at the wrong commit. What kind of review are you looking for here?

dpaez: this kind of review helps, thanks. Gonna check if the tag points to the right place after merge this. The thing is now I can't bump version and prepare the changelog for release from main (can not push from main). So if this doesn't work, I'm going to see if releasing from the ci could help.

juliangruber: > The thing is now I can't bump version and prepare the changelog for release from main (can not push from main)

For this reason usually push to main is allowed for admins. At least that is the case for most repos I know.

PR #194, opened 2020-07-21
Fix: number version metadata

dpaez:
Always cast version to Number


dpaez:
i try to write some tests, but now we are forcing version to be Numbers all the time, so I'm finding difficult to pass a string and get an error, even try to work with `saveItem` method directly

juliangruber:
> but now we are forcing version to be Numbers all the time, so I'm finding difficult to pass a string and get an error

You could revert the fix, then add tests, then commit the fix

dpaez:
Added new test, that is throwing with invalid metadata. Next going to re-apply previous fix.

dpaez:
...And we are good to go now, adding those fixes remind of this classic scene 😅 https://www.youtube.com/watch?v=AbSehcT19u0

juliangruber:
> ...And we are good to go now, adding those fixes remind of this classic scene 😅 https://www.youtube.com/watch?v=AbSehcT19u0

Hahahah a classic! :D

Issue #195, opened 2020-07-21
Inconsistent data is returned from .get()


juliangruber:
1. Have a module, with say `version=58, main=README.md`
2. Call `await p2p.set({ url, main: 'CONTRIBUTING.md })`
3. Call `await p2p.get(url)`, returns `version=58, main=CONTRIBUTING.md`, instead of `version=59, main=CONTRIBUTING.md

This leads to rendering inconsistencies in Hypergraph, discovered in https://github.com/hypergraph-xyz/desktop/issues/69

dpaez:
This should have been closed with #197. We can re-open if it happens again.

PR #197, opened 2020-07-24
Fix: re start related issue

dpaez:
This PR adds a check for only importing writable drives when we do call `refreshMTimes` (this happens on `ready`). It also adds a test that showcase an scenario where this situation could en up with an error. 

Related to https://github.com/hypergraph-xyz/desktop/issues/189


index.js
```diff
@@ -487,6 +489,15 @@ class SDK extends EventEmitter {
         }
 
         // update index.json file
+
+        const dwldHandle = dat.downloadFiles(drive, moduleDir)
+        await new Promise(resolve => {
+          dwldHandle.on('put-end', src => {
+            if (src.name.includes('index.json')) return resolve()
+          })
```
juliangruber, line 497: this would be more native async/await (if it works):

```js
const { once } = require('events')

let src
while (src = await once(dwldHandle, 'put-end')) {
  if (src.name.includes('index.json') break
}
```

index.js
```diff
@@ -487,6 +489,15 @@ class SDK extends EventEmitter {
         }
 
         // update index.json file
+
+        const dwldHandle = dat.downloadFiles(drive, moduleDir)
+        await new Promise(resolve => {
+          dwldHandle.on('put-end', src => {
```
juliangruber, line 495: you're not removing the event listener afterwards

index.js
```diff
@@ -487,6 +489,15 @@ class SDK extends EventEmitter {
         }
 
         // update index.json file
+
+        const dwldHandle = dat.downloadFiles(drive, moduleDir)
```
juliangruber, line 493: why is this called `dwld`?

index.js
dpaez: good catch! 🙈

index.js
dpaez: would you like more something like `downdloadHandle`, my intention was to use an abbreviation.

jameslibscie: I'm personally a fan of longer and more complete variable and parameter names.

juliangruber: Ah! I guess I've never seen download as `dwld` but more `dl`. What about `dlHandle`?

dpaez: `dlHandle` works for me! what do you think @jameslibscie ?

jameslibscie: I'd say `dl` is a common enough abbreviation to be clear to whoever's working on this code :)

dpaez:
sorry that this wasn't updated, find some issues with refresh modification times function.

dpaez:
Note: after rebasing I've found some issues with some of the tests (registering and deregistering, clone related most of them). While trying to fix/update those, the need for having more checks around the main file was needed so I went ahead and added the feature "waiting for the main file" (related to https://github.com/p2pcommons/sdk-js/issues/195).

lib/validate.js
```diff
@@ -726,6 +726,7 @@ const _validateMainExists = async (
   if (indexMetadata.p2pcommons.main.length > 0) {
     const path = join(p2pcommonsDir, key, indexMetadata.p2pcommons.main)
 
+    debugger
```
juliangruber, line 729: debugger

dpaez: 😅

dpaez: removed

PR #206, opened 2020-08-06
Fix: clone watch update

### LINKED ISSUE #187: Local module isn't updated after remote module changes
## Steps to reproduce

1. On machine A, create a profile, and copy that url
2. On machine B, `p2p.clone()` that module
3. On machine A, update the module
4. On machine B, `p2p.clone()` that module _again_

The old module version will be returned by `p2p.clone()`.

dpaez:
This PR improves the tracking of external (cloned) updates.

Closes #187


tests/index.js
```diff
@@ -1765,6 +1765,57 @@ test('cancel clone', async t => {
...
+    dhtBootstrap
+  })
+
+  const p2p2 = new SDK({
+    persist: true,
+    baseDir: dir2,
+    dhtBootstrap
+  })
+
+  await p2p2.ready()
```
juliangruber, line 1784: this call isn't necessary, right?

dpaez: right, wasn't needed

PR #208, opened 2020-08-12
Key location update

chartgerink:
After merging of https://github.com/p2pcommons/specs/pull/49, this PR will make the `sdk-js` up to spec.

Note: It currently was out of spec given that the key folder had been changed to `.hyperdrive` already.


chartgerink:
Yea definitely breaking. That's why I did it now given that we still can. We don't need to copy at this moment, we can still do resets.

This should be the final breaking change before beta.

PR #210, opened 2020-08-12
Update: read only modules

### LINKED ISSUE #172: Read-only registered modules
**Is your feature request related to a problem? Please describe.**
Registered content is supposed to be immutable. I can now still edit it on my local version.

**Describe the solution you'd like**
To prevent people from messing around with registered content on their local device, it helps to lock versioned folders to read-only by default.

**Describe alternatives you've considered**
-

**Additional context**
-

dpaez:
This PR introduces `read-only` permission only for **versioned** modules.  

Closes #172


dpaez:
Looks like its failing only on windows. Would need to dig more into this.

chartgerink:
Code looks good to me from my reading levels 😊 

Here's the error from the log, which isn't particularly helpful either.

```
    × cloned module is writable
    ----------------------------
      operator: fail
      at: Test.<anonymous> (D:\a\sdk-js\sdk-js\tests\index.js:1769:7)
      stack: |-
  Error: cloned module is writable
  at Test.assert [as _assert] (D:\a\sdk-js\sdk-js\node_modules\tape\lib\test.js:228:54)
  at Test.bound [as _assert] (D:\a\sdk-js\sdk-js\node_modules\tape\lib\test.js:80:32)
  at Test.fail (D:\a\sdk-js\sdk-js\node_modules\tape\lib\test.js:322:10)
  at Test.bound [as fail] (D:\a\sdk-js\sdk-js\node_modules\tape\lib\test.js:80:32)
  at Test.<anonymous> (D:\a\sdk-js\sdk-js\tests\index.js:1769:7)
```

Might it be that because Windows is not UNIX based it has different file ownership rules and `chmodr` doesn't work the same there? (long shot)

README.md
```diff
@@ -190,6 +190,8 @@ Returns a boolean indicating whether the module is verified.
 
 Get a module from the local db or the swarm. If the module is not present on the local db and the swarm is enabled (`disableSwarm === false`) then it will query the swarm.
 
+:warning: NOTE: **versioned** modules directories are `read-only`.
```
juliangruber, line 193:
```suggestion
:warning: NOTE: **versioned** module directories are `read-only`.
```

double plural

index.js
```diff
@@ -492,6 +496,11 @@ class SDK extends EventEmitter {
 
         // update index.json file
 
+        // make moduleDir writable first
+        if (!drive.isCheckout) {
+          await this.makeWrite(moduleDir)
+        }
```
juliangruber, line 502: what if any error happens afterwards? should we use a try/catch/finally here, so that even if an error happens it will be made read only afterwards?

index.js
```diff
@@ -1194,6 +1208,75 @@ class SDK extends EventEmitter {
...
 
+  async makeWrite (moduleDir) {
+    assert(typeof moduleDir === 'string', 'moduleDir should be a path')
+    debug('makeWrite: %s', moduleDir)
+    const mode = 0o777 // write, read and execute
+    try {
+      await access(moduleDir)
+    } catch (err) {
+      this._log(err.message, 'warn')
+      return
```
juliangruber, line 1219: why abort here? if the dir isn't accessible, that's an error, right?

index.js
```diff
@@ -1194,6 +1208,75 @@ class SDK extends EventEmitter {
...
+    }
+    const stat = await statFn(moduleDir)
+    const permString = '0' + (stat.mode & 0o777).toString(8)
```
juliangruber, line 1222: could you add a comment here?

index.js
```diff
@@ -492,6 +496,11 @@ class SDK extends EventEmitter {
 
         // update index.json file
 
+        // make moduleDir writable first
```
juliangruber, line 499: I'm not sure I follow the logic. In my understanding, the dir `${key}` is always writable, and the dirs `${key}+${version}` are never writable. Isn't that true?

index.js
```diff
@@ -1194,6 +1208,75 @@ class SDK extends EventEmitter {
...
+      this._log(err.message, 'error')
+      // throw err
+    }
+  }
+
+  async makeReadOnly (moduleDir) {
+    assert(typeof moduleDir === 'string', 'moduleDir should be a path')
+    const mode = 0o555 // read and execute
+    const stat = await statFn(moduleDir)
+    const permString = stat.mode & 0o777
```
juliangruber, line 1241: could you add a comment for this too?

index.js
chartgerink: `${key}` might not be writable if the Hyperdrive is somebody else's?

juliangruber: ah gotcha, but the PR description says `This PR introduces read-only permission only for versioned modules.`

chartgerink: Oh sorry! You're right 🙈 Also part of my own comment https://github.com/p2pcommons/sdk-js/issues/172#issuecomment-671973715

juliangruber (changes requested):
needs a rebase

dpaez:
rebased cc @juliangruber

juliangruber:
the files changed tab still shows changes from another PR that aren't yet in master :/

dpaez:
> the files changed tab still shows changes from another PR that aren't yet in master :/

I double checked the Commits and Files changed tabs and everything looks related to this update. Maybe I'm a bit lost here.

dpaez:
I think this needs a new rebase now 😓

dpaez:
Rebased and fixed! :)

index.js
```diff
@@ -492,13 +496,32 @@ class SDK extends EventEmitter {
...
 
-        // If an external drive have changed, update it locally
-        const dlHandle = dat.downloadFiles(drive, moduleDir)
-        debug('refreshMTimes: waiting for index.json...')
-        await dlWaitForFile(dlHandle, join(moduleDir, 'index.json'))
+        let dlHandle
+        try {
+          // make moduleDir writable first
+          if (!drive.isCheckout) {
+            await this.makeWrite(moduleDir)
```
juliangruber, line 503: maybe `makeWritable` is more obvious to the reader

index.js
```diff
@@ -492,13 +496,32 @@ class SDK extends EventEmitter {
...
-        this.dls.set(dkey, dlHandle)
+          const dkey = DatEncoding.encode(drive.discoveryKey)
+          this.dls.set(dkey, dlHandle)
+        } catch (err) {
+          this._log(err.msg, 'warn')
+        } finally {
+          // make it read only again
+          if (drive.isCheckout) {
+            if (dlHandle) {
+              dlHandle.on('end', () => this.makeReadOnly(moduleDir))
```
juliangruber, line 519: errors thrown from `this.makeReadOnly` can't be caught, what about using the `once` function again?

index.js
```diff
@@ -1209,6 +1232,80 @@ class SDK extends EventEmitter {
...
   }
 
+  async makeWrite (moduleDir) {
+    assert(typeof moduleDir === 'string', 'moduleDir should be a path')
+    debug('makeWrite: %s', moduleDir)
+    const mode = 0o777 // write, read and execute
+    try {
+      await access(moduleDir)
+    } catch (err) {
+      this._log(err.message, 'warn')
```
juliangruber, line 1242: why log the error if we also throw it? I think this try/catch could just be removed

index.js
```diff
@@ -1209,6 +1232,80 @@ class SDK extends EventEmitter {
...
+      this._log('module directory is already writable', 'warn')
+      return
+    }
+
+    try {
+      await chmodr(moduleDir, mode)
+    } catch (err) {
+      this._log(err.message, 'error')
+      throw err
+    }
```
juliangruber, line 1261: same here as above

index.js
```diff
@@ -1209,6 +1232,80 @@ class SDK extends EventEmitter {
...
+   * @returns {boolean} - Indicates if the module directory has been created (true) or not (false)
+   */
+  async createModuleDir (moduleDir) {
+    assert(typeof moduleDir === 'string', 'moduleDir is required')
+    debug('createModuleDir: creating module dir %s', moduleDir)
+    try {
+      await mkdir(moduleDir)
+      return true
+    } catch (err) {
+      if (err.code === 'EEXIST' || err.code === 'EACCES') {
```
juliangruber, line 1300:
```suggestion
      if (err.code === 'EEXIST' || err.code === 'EACCESS') {
```

index.js
juliangruber: To me that is fine as it is :)

index.js
```diff
@@ -492,13 +496,32 @@ class SDK extends EventEmitter {
...
-        // If an external drive have changed, update it locally
-        const dlHandle = dat.downloadFiles(drive, moduleDir)
-        debug('refreshMTimes: waiting for index.json...')
-        await dlWaitForFile(dlHandle, join(moduleDir, 'index.json'))
+        let dlHandle
+        try {
+          // make moduleDir writable first
+          if (!drive.isCheckout) {
+            await this.makeWritable(moduleDir)
+          }
```
juliangruber, line 504: The module dir is made writable if it's not a checkout, and read only only if it is a checkout. Doesn't that mean that if it's not a check out, it doesn't need to be made writable, since it never was made read only?

index.js
```diff
@@ -1209,6 +1232,82 @@ class SDK extends EventEmitter {
...
+   * @description Performs a chmodr (recursive) on a module directory, making it writable (mode: 777)
+   * @private
+   * @async
+   * @param {string} moduleDir -  path to module directory
+   */
+  async makeWritable (moduleDir) {
+    assert(typeof moduleDir === 'string', 'moduleDir should be a path')
+    debug('makeWritable: %s', moduleDir)
+    const mode = 0o777 // write, read and execute
+    await access(moduleDir)
```
juliangruber, line 1247: Why the `access` call here? If it's not accessible, the `statFn` or `chmodr` below would throw anyways

juliangruber (changes requested):

juliangruber (approved):

Issue #220, opened 2020-08-19
listContents() doesn't reliably return latest version


jameslibscie:
> Using some `console.log`s, I noticed that the same module is returned with different versions in two consecutive `p2p.listContents()` calls:
> 
> First call, where it's listed as draft (`version = 11`)
> 
> ![image](https://user-images.githubusercontent.com/10247/90646441-80311380-e237-11ea-8a85-58a6b19fd2c5.png)
> 
> Second call, where it isn't (`version = 4`)
> 
> ![image](https://user-images.githubusercontent.com/10247/90646459-8626f480-e237-11ea-8680-409aab8491d0.png)

This causes modules to display twice in Hypergraph, for example 1 draft version and 1 registered version.

dpaez:
I think that somehow we entered wrong/bad data to the localdb to reach a case like the one described in the issue. 

We can add some checks on `listContent` method to filter values and return the one with the latest modified time or latest version. But which one? It should be the same.

On the other hand, p2pcommons/sdk-js#221 is an attempt to prevent this.

jameslibscie:
Should be fixed with #221

PR #221, opened 2020-08-19
fix: saveItem

dpaez:
This PR is an attempt to improve a possible scenario where we store in the localdb a mismatch between `version` and `lastModified` values. i.e.: version is bigger than other module but the lastModified is older. Both values should go bigger with time.

This could help prevent things like #220


jameslibscie:
@juliangruber Could you perhaps test whether this solves the problem?

juliangruber (changes requested):
The same issue unfortunately

image

image

dpaez:
@juliangruber and @jameslibscie I've updated the `list content` test with some `set` operations in between and more comparisons. 

Is this similar to the behaviour that ends with the bug?

juliangruber:
The steps in Hypergraph are

1) Create profile
2) Create content
3) Register content
4) _Restart the app_
5) List contents

dpaez:
Thanks @juliangruber this is super helpful

juliangruber:
Sorry for not providing these steps earlier 😅

dpaez:
Added a new test that follows the steps that @juliangruber described before. It's passing so maybe this is an indication that the issue is somewhere? Do we have any other info to add here?

juliangruber:
> Added a new test that follows the steps that @juliangruber described before. It's passing so maybe this is an indication that the issue is somewhere? Do we have any other info to add here?

No other info besides using the steps above you should be able to reproduce this in Hypergraph Desktop:

1) Reset DB
1) Create profile
1) Create registered module
1) Restart app
1) Open drafts -> you see the registered module
1) Navigate somewhere else
1) Open drafts -> you don't see the registered module

I looked at the implementation of `listContents` and saw that it just returns the DB items. That makes me wonder how after launching Hypergraph, where we _don't_ write anything to the DB, only read, the returned values could be different.

Is there maybe something happening in `this.ready()` that modifies the DB items?

juliangruber:
> Yes. On `ready` we call the `refreshMTimes` that is used to check if something changed while the sdk was offline. This function can update the localdb if that happens.

Thanks, I'm about to try out these new changes. I still have no clue why during or right after `refreshMTimes` the version would be higher than afterwards though

juliangruber (approved):
This seems to have fixed it Diego :O <3

I can't reproduce any more :)

jameslibscie (approved):
All good!

dpaez:
Glad it worked. I think that we can go with this fix for now. I'd like to recheck the `refreshMTimes` function soon.

PR #222, opened 2020-08-19
Update: delete versioned modules

### LINKED ISSUE #215: Delete versioned modules
As discussed in https://github.com/hypergraph-xyz/desktop/issues/219#issuecomment-674828962, we currently only have a method to delete unversioned modules. 

This means that when I delete a module, we currently *cannot* delete the registered version itself.

Could we add an optional argument for the module version? We would need to check whether that is actually in the database, and then delete if it is.

dpaez:
This PR enables `delete` method to receive not only _unversioned_ urls but also versioned.

Closes #215


chartgerink (approved):
Looks good from my limited reading skills 😊

PR #224, opened 2020-08-20
Update: restore export

### LINKED ISSUE #211: Export p2pcommons method
A method to export p2pcommons for easy restore

* the private keys
* `settings.json`

The method should:
- [ ] Collect the private keys
- [ ] Collect `settings.json`
- [ ] Take a path to write the export to
- [ ] Write out the export file

dpaez:
New feature: export method

This PR closes #211


README.md
```diff
@@ -264,6 +264,12 @@ If `deleteFiles` option is true, then the target folder will be moved to the tra
 
 Closes the swarm instance (if created) and the local db.
 
+### export
+
+> _async_ `export(exportPath: string)`
+
+Collect private keys and `settings.json` file and saves them in the received path (`exportPath`)
```
juliangruber, line 271: What about local modules that haven't been shared with the network yet? Or is the plan for these to be backed up manually?

chartgerink: This is a good question :) 

The practical consequence will be that the drafts would only be restored up through the backup, which would be helpful in (1) device migration but less for (2) loss of devices. 

Do you think we would need this now? This export is primarily aimed to cover (2) at the moment, I would suggest, with (1) following in a while when we expand on the pattern for this to manage expectations.

chartgerink:
Closing this as its a stale PR. Will be available in the archive if we decide to re-implement this at some point.

PR #225, opened 2020-08-25
Release: bump version and changelog

dpaez:
New candidate release: **0.8.0** 🎉 
This PR bumps package.json and updates changelog accordingly.


jameslibscie (approved):
Tested with Hypergraph 0.6.0 and this seems OK. Delete doesn't work properly, but that is expected due to change in behaviour of the `delete()` method. @juliangruber this will need updating in Hypergraph probably. Would like to explore some different delete scenarios of versioned and unversioned, registered and unregistered content when the time comes since there are quite a few combinations here that might need specific behaviour.

jameslibscie:
@dpaez @juliangruber Upon database reset, it fails to delete versioned folders on Windows, but it can delete the files in it.

jameslibscie:
Also just as a note, the read-only on Windows only applies to existing files and only applies to updating those files. I can delete the files without any warning and I can add new files that aren't read-only. I can also delete the entire folder.

jameslibscie:
Closing because @dpaez and I discovered that `delete()` with an unversioned key should delete all versions. This is the most realistic use case and should fix the issues Hypergraph currently has with deleting. `delete()` with a versioned key should just delete that specific version.

The release will wait until that fix is included.

PR #227, opened 2020-08-25
Update: delete remove all versions with unversioned key

### LINKED ISSUE #232: Delete tries to deregister from remote profile
## Description

When deleting a content module that is has local _and_ remote authors, and it has been registered to the local author, the SDK will also try to deregister it from the remote author, whose profile is not writable.

## To Reproduce

1. Init local profile
2. Init remote profile
3. Follow remote profile
4. Init content module with both as authors
5. Register content module to local profile
6. Delete content module

## Expected behavior

The content module is removed from the local profile but the remote profile stays untouched.

## Actual behavior

The SDK tries to remove the content module from the remote profile and throws an error that the profile is unwritable.

## Additional context

Even if the remote profile has the content module registered, it wouldn't be writable, so delete shouldn't try to modify it.

dpaez:
This PR tweaks the way delete was working.

It contain the following changes:

- From https://github.com/p2pcommons/sdk-js/pull/227#issuecomment-683806853:
  - Receives unversioned key (otherwise error)
  - Deregisters all versioned and unversioned registrations from profile
  - Removes from DB
  - Deletes all versioned and unversioned folders from disk
- Only deregister writable profiles. Fixes #232


index.js
```diff
@@ -1872,6 +1874,12 @@ class SDK extends EventEmitter {
       )
 
       if (deleteFiles) {
+        try {
+          await access(drivePath)
+        } catch (err) {
```
juliangruber, line 1879: this could also throw if the directory exists but permissions forbid it to be accessed

juliangruber: I just looked at the trash module's source code, and it's using synchronous file system calls.

Also it looks like the trash module already does a check like this:

https://github.com/sindresorhus/trash/blob/d6d2c3a86125925b4e04749a8e3a311af48e93b8/index.js#L34-L39

juliangruber: Do you know where the process hangs exactly?

juliangruber (changes requested):

index.js
dpaez: Hi juli :wave: Yeah, I recognise this looks like an extra thing to do. 
We saw this issue yesterday when trying hypergraph with the latest sdk that you can get from master. 

1. you create your profile
2. create and register content
3. try to delete the content

What happens next is, first time you call delete, it succeeds quickly, but you still see the content attached to the profile. So if you hit delete again, it will hang. 
I put a debugger in hypergraph right after this (https://github.com/hypergraph-xyz/desktop/blob/fe1ce6de64bf4ebafd1d8367342e4711b89f9391/app/components/content/content.js#L353) and it never reaches that point. I try the same operation with this little fix and it worked.

You are right about the permissions, If we add these line it should be moved right before the trash call. Because in the middle we already address the permission check.

juliangruber: When I try to delete a module with latest sdk-js, I see this:

image

dpaez: I'm not able to reproduce this. The only thing similar I can reproduce is, if I try to `delete` registed content and the version is not passed as part of the key to the delete method then `delete` is trying to remove the unversioned directory. And that works. But the profile is still pointing to the versioned directory. So, on the next restart of the sdk/hypergraph you will see the content there again.

jameslibscie:
@juliangruber, @dpaez and I decided that the best way to go would be as follows:
- Receives unversioned keys (otherwise error)
- Deregisters all versioned and unversioned registrations from profile
- Removes from DB
- Deletes all versioned and unversioned folders from disk

juliangruber (approved):
It works!

dpaez:
@juliangruber and @jameslibscie I've updated the PR with the deregister fixes.

index.js
dpaez: ended up removing the access call.

PR #229, opened 2020-08-26
Update: new manual and auto sync 

### LINKED ISSUE #228: listContent doesn't show updated version after file updates
## Description

If I update files for a content module and then run `listContent()`, it doesn't show an updated version. This means that Hypergraph can't register the module at the correct version, which will then lead to users thinking data has been lost. Perhaps a new version isn't even created.

## To Reproduce

1. Create content module with at least one file
2. Run `listContent()` -> see version
3. Update the file
4. Run `listContent()` again -> see version is the same

## Expected behavior

A new version is created immediately and `listContent()` shows that version

## Actual behavior

No new version is created until an update is triggered manually by the SDK

## System

- OS (incl. version): Windows 10
- SDK version: 0.8.0 candidate #225

dpaez:
This PR attempts to fix #228 by improving how we are checking the file existence condition. Instead of just checking if there is a file with the passed _name_ we now compare file and drive stat's **mtime** for that file. This should be more robust.

Closes #228


jameslibscie:
The bug with renamed main file still exists and I've verified that it's the same cause as #228 

1. I create a new module with a file that is the main file
2. I edit that file
3. I use `get()` or `listContents()`, it returns version 5, even though I've updated a file -> this is the root cause
4. I rename the file
5. I use `get()` or `listContents()`, it returns version 5, even though I've renamed the file -> same problem again
6. I use `set()` to update the main file, but it fails, looking for the *previous* main file, which no longer exists -> new problem, which may be caused by it somehow verifying the old metadata? I'm actually not so sure about this anymore.

Either way, #228 is more about making sure that the SDK bumps the version number as soon as any file in the drive changes, as long as Hypergraph is running of course. Otherwise, we can't really reliably allow file editing at all...

jameslibscie:
If we're having a hard time getting the file changes to be seen automatically, would a good workaround perhaps be to have a `refresh(key)` method that rescans the drive, updates the DB and can be awaited before performing certain actions like register?

dpaez:
Quick update: I've been toying around an implementation that will update the local db on every 'put-end' event that is emitted by the mirror-folder watching that folder. I'm getting some false positives so I need to improve and maybe filtering some updates. 

I don't discard the possibility of having a new `sync` method also. This would work better. So I probably end up adding support for both strategies maybe or just the `sync`.

dpaez:
Added a new `sync` method AND make use of some local hooks to auto update on changes. So its not necessary to call `sync`, but in some cases, like if the user needs to do an operation that needs the latest version then calling `sync`. Eg: user modifies content, like add new files to a content module. Then immediately after wants to register that content module. In that case, having the latest version is important. We can hook and listen to the `update-content` event or just call `sync`. 

A side effect is that we are possibly incrementing the version more than once by overwriting the drive. I'm looking if there is an easy/realiable way of avoiding/filtering changes.

jameslibscie:
The versions update immediately - this is working great! Except I can't seem to get `sync()` to work a lot of the time:

image

Doesn't matter whether I `encode()` the URL or not. It did work at first though!

jameslibscie:
I was thinking about the name of `sync()` and think it might be confusing as it might suggest syncing with the network. How about `refreshDrive()`?

dpaez:
> I was thinking about the name of `sync()` and think it might be confusing as it might suggest syncing with the network. How about `refreshDrive()`?

Yeah, I like it. Also `syncDrive` and `updateDrive`.

README.md
```diff
@@ -149,6 +150,12 @@ Used to update an existing module.
 
 - metadata: object with updated or added values. The only required field is the `url` property, which is used to identify the module to update.
 
+### sync
+
+> _async_ `sync(key: string|buffer, moduleDir: string, opts: object)`
```
juliangruber, line 155: What about instead of passing the `dir` we pass `key` and optionally `version`? The SDK knows enough to construct the `dir` from that.

README.md
```diff
@@ -149,6 +150,12 @@ Used to update an existing module.
...
+
+Synchronize an hyperdrive with a module directory. Optionally accepts an opts object. These are passed to [dft](https://github.com/pfrazee/diff-file-tree).
```
juliangruber, line 157: It would be useful to explain some more why synchronizing helps. For example, when do you need to synchronize it?

index.js
```diff
@@ -632,6 +636,51 @@ class SDK extends EventEmitter {
...
+      }
+      await this.localdb.put(DatEncoding.encode(moduleUrl), {
+        ...metadata,
+        rawJSON: indexJSON,
+        avroType: this._getAvroType(indexJSON.p2pcommons.type).name
+      })
+
+      // emit update-profile|content event
+      this.emit(`update-${indexJSON.p2pcommons.type}`, file.name)
+    } catch (err) {
```
juliangruber, line 679: This try/catch makes the method decide how to handle errors, not the consumer. What about removing it here and changing the call below to

```js
this._updateLocalDB(publicKeyString, dst).catch(err => this._log(err.message, 'error'))
```

This way to me it would be clearer when reading the code.

index.js
```diff
@@ -1049,6 +1105,56 @@ class SDK extends EventEmitter {
...
+      // dat open
+      drive = dat.open(this.store, DatEncoding.decode(key), {
+        sparse: true,
+        sparseMetadata: true
+      })
+
+      await drive.ready()
+    }
+    if (!drive.writable) {
+      this._log('drive is not writable -  can not sync')
```
juliangruber, line 1134:
```suggestion
      this._log('drive is not writable -  nothing to sync')
```

This makes it clearer that it's not an error at all since nothing is to be done

juliangruber (changes requested):

juliangruber:
I think this PR's name could be updated because it's now about the `sync` method

README.md
jameslibscie: `sync` only makes sense on the unversioned folder as the versioned folder shouldn't be updated and wouldn't trigger a new version. I agree that it should only be the key passed.

README.md
jameslibscie:
```suggestion
Synchronize a hyperdrive with a module directory to ensure that the sdk is aware of the latest changes to files. Optionally accepts an opts object. These are passed to [dft](https://github.com/pfrazee/diff-file-tree).
```

README.md
dpaez: going with the `key` as the only parameter

index.js
dpaez: I like it!

juliangruber (approved):
Looks great!

jameslibscie (changes requested):
Haven't tested, but would strongly suggest renaming `sync()` to something else

dpaez:
Renamed sync method to `refreshDrive`

juliangruber (approved):

jameslibscie (approved):
Awesome!

Issue #231, opened 2020-08-31
False positives on main file extension check


jameslibscie:
The main file extension check currently disallows `.js` files. Since there are likely many plain text file types that are not on the list, it might be wise to implement a plain text checker, like:

[https://www.npmjs.com/package/istextorbinary](https://www.npmjs.com/package/istextorbinary)
[https://www.npmjs.com/package/isbinaryfile](https://www.npmjs.com/package/isbinaryfile)
[https://www.npmjs.com/package/chardet](https://www.npmjs.com/package/chardet)

See also: https://www.notion.so/File-spec-validation-plan-a059300d54c546199b6e08f75b542a2f

dpaez:
Closed with #233

PR #233, opened 2020-09-01
Add plain text validation

chartgerink:
Added a first implementation after talking with @dpaez for issue #231. Tried adding a test but wasn't succesful at the moment.


dpaez:
@chartgerink updated! I think is good to work

lib/validate.js
```diff
@@ -707,13 +708,19 @@ const _validateMainRelativePath = indexMetadata => {
...
     return ext[ext.length - 1]
   }
 
   const ext = getExtension(indexMetadata.p2pcommons.main)
+  const mainPath = join(p2pcommonsDir, key, indexMetadata.p2pcommons.main)
+
+  let mainBuffer = null
+  try {
+    mainBuffer = await readFile(mainPath)
+  } catch (_) {}
```
juliangruber, line 723: what's the reason for ignoring errors here?

juliangruber (changes requested):

chartgerink:
Nice, thanks @dpaez! I see I was being too hasty by adding the path without concatenating the p2pcommons dir. The test needed a file written out I see as well, thanks for teaching me

lib/validate.js
dpaez: This code is part of `_validateMainExtension` function which is part of `validateMain` function. The `_validateMainExtension` runs after `_validateMainExists`, a function that checks for the existence of the main file. So we can assume that a main file should exist up to this point.

I think I wanted to be protected if the function run outside `validateMain` and/or something weird happens with readFile fn. This is connected to the fact that we are adding this double check like the extension check and the content one. If there is no `mainBuffer` for some reason, then the content check wont happen and we will end up with two possibly different validations on the extensions of the main file.  So the absence of the mainBuffer does not break things. Unless we strongly wanted to have the content validation.

dpaez: I put some seconds thoughts on this and I think it's safe to remove that try/catch cage. Release the readFile! :)

juliangruber: I think we kind of have to do content validation, if "it's plain text" makes it valid. Good point about _validateMainExtension being callable outside of the validateMain function, I assumed it would be private because it starts the an underscore.

juliangruber (changes requested):

lib/validate.js
```diff
@@ -663,7 +664,7 @@ const validateMain = (exports.validateMain = async ({
   _validateMainNoDotfile(indexMetadata)
   _validateMainRelativePath(indexMetadata)
   await _validateMainExists(indexMetadata, key, p2pcommonsDir, _flat)
-  _validateMainExtension(indexMetadata)
+  await _validateMainExtension(indexMetadata, key, p2pcommonsDir)
```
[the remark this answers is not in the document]
dpaez: I'll go with `_validateMainContentType`

PR #238, opened 2020-09-07
Update: tweak edit content test

dpaez:
Use `version` and or `lastModified` to know if drive has been updated


lib/utils.js
```diff
@@ -174,7 +174,7 @@ exports.dlWaitForFile = async (dlHandle, driveFileStatMtime, file) => {
   file = normalize(file)
   try {
     const fileStat = await stat(file)
-    if (fileStat.mtime.getTime() < driveFileStatMtime.getTime()) {
+    if (fileStat.mtime.getTime() <= driveFileStatMtime.getTime()) {
       throw new Error('file in filesystem is older than the one in the drive')
```
chartgerink, line 178:
```suggestion
      throw new Error('file in filesystem is the same or older than the one in the drive')
```

Out of curiosity, what led to this change?

chartgerink (approved):
I left a minor suggestion in the error code, but nothing else from my side :)

lib/utils.js
dpaez: > Out of curiosity, what led to this change?

This PR is related to a previous one (#237). From time to time, when we run the tests on the ci they freeze and I had to cancel the job and run it again or it will be cancelled automatically after a few hours. I think there were some async race conditions in some cases (some tests need to actively wait for an event or that certain situation happen to continue). So #237 and this PR are tweaks and some cleaning up to try to make tests more stable again.

PR #249, opened 2020-09-25
Fix refreshMTimes() hangs on unavailable modules. Closes #248

### LINKED ISSUE #248: ready() hangs on drive.stat('/')
There's an issue with Hypergraph where sometimes the app won't load, until you reset the database. I've now seen this on my machine as well and added some debug statements.

During `this.ready()` it calls `refreshMTimes()`, which iterates over all modules in series. The call where it hangs is

https://github.com/p2pcommons/sdk-js/blob/bf27cee69f3881e451d7be54a2bfbdccebabd575/index.js#L444

I assume this could mean that the drive hasn't been downloaded yet, but already is in the db. Could that be?

If so, I'd see two ways to solve this:

1) Only add drives to the db once `drive.stat('/')` works
2) Add a timeout to `refreshMTimes()` so that individual modules will be ignored

juliangruber:
If `drive.stat()` doesn't return in the same tick, the module isn't yet available.

Closes #248.


dpaez (approved):
This is great!

dpaez:
@juliangruber Related to the failing test, what if we modify the timeout to something small too, like `100`ms just to do a little wait for the stats. I'm seeing that this helps the tests to complete OK, at least that one which is checking using a local update. Makes sense? 🤔

juliangruber:
If that fixes it I'm all for it

PR #252, opened 2020-09-29
Update: add download info, emit new download events, resume download

dpaez:
This PR adds:
- new method `getDownloadInfo`
- new events: `download-resume`, `download-started`, `download-progress` and `download-complete`.
- resumes download on refreshMTimes


dpaez:
A few notes about this PR:

> main goal: "re-enable downloads when the SDK re-starts"

- This PR works around `refreshMTimes` function at sdk bootstrap. We call a new method called `resumeDownload` that uses the contentFeed to resume download if needed. 

- `resumesDownload` downloads everything into the hyperdrive, when that is complete we call `refreshFS` to update the FS with the changes. So this is a thing, **the user wont be able to see the updates until is complete**. But the resume seems to be working, I was able to log some `dlInfo` objects and saw their downloaded value increase.

- There is a new value in the metadata: `isCheckout`. I noticed that I was needing this value and it turns out you can not ask a drive if it is a checkout, after you create it. Long story short, we know when we are dealing with a checkout right after we call the `.checkout(version)` on that drive. **This results in a db reset, since it is incompatible**. Sorry about this.

- clone method returns a new value: `{dlInfo}` with download info and resume and cancel methods.

- Finally we have events to track progress of things: 
  - `download-resume`: to know a resume is happening
  - `download-started`: for new downloads!
  - `download-progress`: to know the progress of `key`
  - `download-complete`: to know that is ready! 

All of this events come with an object: `{key, percentage}`

:v:

chartgerink:
Thanks @dpaez! 

Is there a way for us to include a database migration tool in the `sdk-js`? Or a way to repopulate the database without losing the private keys?

After yesterday, database resets are no longer feasible. The first content is being shared through Hypergraph/p2pcommons. If a database is reset the private keys are as well, and given that we don't have a way to restore those it means we're creating a breaking change that people cannot recover from.

juliangruber:
A migration is not possible, because the `isCheckout` information wasn't stored anywhere. It can only be stored right when the drive is created, but not deducted from its current state.

juliangruber:
> After yesterday, database resets are no longer feasible. The first content is being shared through Hypergraph/p2pcommons

Couldn't we updated the 2 profiles and 1 content module today still? The SDK isn't at 1.0.0 yet so it wasn't communicated that breaking changes are no longer possible, as far as I'm aware?

juliangruber:
Or does that mean @chartgerink you shared production data a day before release?

chartgerink:
The first piece of content I posted yesterday with a fellow researcher and is in the Vault, is linked to our private keys. The database itself is not my issue at the moment, but the implication of a reset right now is also resetting the private keys. 

In a sense I used it for production before the release, yes, because it was part of the prep for the release. I could probably repost it today, if after that we know 100% it'll stay stable. Private key retainment is what I'm worried about here. 

Pinging you on mattermost to chat and will reportback here for documentation :)

chartgerink:
@juliangruber is looking into whether we can compartmentalize the reset of the database and the reset of the private keys. 💯

juliangruber:
It looks like the progress events arrive somewhat randomly:

image

Also while this is happening the files' sizes on disk don't increase.

dpaez:
> It looks like the progress events arrive somewhat randomly:
> 
> image
> 
> Also while this is happening the files' sizes on disk don't increase.

I think this is related to the non-linear download of every block. The percentage value its something that we artificially create and maybe its not the best, given this situation, unless we enable the linear download. The download event is giving us an `index` value that correlates to the downloaded block.

Anyway if we use percentages, the linear download needs to be enabled.

juliangruber:
We don't need percentages yet, so that works!

dpaez:
OK, i tested this locally, with one of the tests (see `resume download clone`) and it works. The main change was to emit a new event after the `refreshFS` is done. 
The new event is `download-resume-completed`. It's only emitted when a resumed download finishes OK and after sync it to the fs. 
Also, the previous `download-complete` has changed, now its `download-drive-complete`. I think this reflects better what the event is saying, the drive is completed, we still need to sync it to the fs. Currently mirror-folder is handling that part. The new resume mechanism is handling the resumes (sdk re-starts)

chartgerink:
While testing this PR, I discovered that (on Windows) I could do the following:

1. Start downloading a big module
2. Interrupt the download by CTRL+R
3. The module would remain on disk as it was before the download was interrupted
4. Remove the module folder (`~/.p2pcommons-dev/<hash>+<version>`) manually
5. Go back to the profile of the module's author / feed
6. Open module
7. Open folder 
8. The complete file was there

This was after I waited for the console `download-progress` to finish logging.

When trying to recreate this with @dpaez (he was on macOS) we were not able to replicate on there. I was able to replicate this reliably on my machine.

chartgerink (approved):
The resume functionality currently works silently in `.hyperdrive` using this PR. It does not write out to disk yet, so if a download is interrupted it will resume downloading but that will not be exposed to the end user based on this.

Cheers @dpaez for the lovely vibes and the energy you poured in to create this 💜

dpaez:
Note: we temporarily disabled the `makeReadOnly` of checkout folders.

PR #254, opened 2020-10-01
Revert "Update: add download info, emit new download events, resume download"

chartgerink:
Hi @dpaez 

After some more testing with @juliangruber we were able to get 6fbc85937063257e8d70e81f7f303b3f2dfe2c07 functioning in Hypergraph after all. We now npm linked to this specific branch, but that means that the main on `sdk-js` needs to be reverted.

Posting this draft to work on this sometime after the beta release and for documenting what's going on.

Cheers,
Chris


juliangruber:
> but that means that the main on `sdk-js` needs to be reverted.

It doesn't need to be reverted, the main branch wasn't released and is to be considered a dev version, a work in progress. We can also add commits to fix what is there currently, instead of rolling back.

PR #255, opened 2020-10-13
Fix: resuming downloads + general clone fixes

dpaez:
This PR improves the clone methods and the download resume feature. 
It also adds a new `addFiles` method (and `removeFile`) that it should be used to add new files to a module.


chartgerink (approved):
@dpaez and I tested this last week or so with [Hypergraph](https://github.com/hypergraph-xyz/desktop/pull/312), and it worked quite well. 

What this has learnt us is that it is becoming more important to refactor the `sdk-js` to reduce complexity of the codebase and reduce complexity of these PRs.

PR #257, opened 2020-11-20
Update: upgrade from corestore-swarm-networking to corestore/networker

dpaez:
This PR updates the swarm dependency to its latest.


chartgerink (approved):
Nice, thanks ☺️ 

Double checked with the [corestore-networker update guide](https://github.com/andrewosh/corestore-swarm-networking/blob/master/UPGRADE.md) before approval.

PR #260, opened 2021-01-27
dpaez/add resume download start

chartgerink:
This improves networking, as tested in the implementation of https://github.com/hypergraph-xyz/desktop/pull/312.

Given the time passed since the original branch was created, I'm putting this as a PR and any additional work can be noted in for example the issues or elsewhere.


chartgerink:
I am going to merge this for now. It seems like the code has improved substantially and letting this sit is only going to create rebasing issues in the future. Especially given the refactor I was already doing, which is going to create lots of conflicts or backtracking if this is not merged.

I will work through the tests and make sure all tests can be unskipped (see #262).

PR #265, opened 2021-03-03
Remove chan dependency

chartgerink:
Given that we updated the release procedure to `np`, this PR removes the chan dependency that is no longer used. Commit messages will now be used to provide release notes.


chartgerink:
This build kept failing because of recursion: The chan check was still part of the actions when the PR was created 😅

PR #276, opened 2021-03-03
Refactor saveItem test

chartgerink:
@dpaez I removed the `dhtBootstrap` from this test as I couldn't make sense of why it was needed here? 

I figured the following: We're locally creating a module and validating that, without any network connectivity necessary. 

Please advise ☺️

### DIFF
--- a/tests/test-save-item-method.js
+++ b/tests/test-save-item-method.js
@@ -1,0 +1,58 @@
+const tempy = require('tempy')
+const test = require('tape')
+const SDK = require('..')
+
+module.exports = () => {
+  test('saveItem: should throw ValidationError with invalid metadata', async t => {
+    const dir = tempy.directory()
+
+    const p2p = new SDK({
+      baseDir: dir
+    })
+
+    const sampleData = {
+      type: 'content',
+      title: 'demo',
+      description: 'lorem ipsum'
+    }
+
+    const { rawJSON } = await p2p.init(sampleData)
+
+    const key = rawJSON.url
+
+    try {
+      const { rawJSON: updated, metadata } = await p2p.saveItem({
+        isWritable: false,
+        lastModified: new Date(),
+        version: '5',
+        indexJSON: {
+          url: key,
+          title: 'demo',
+          description: 'something new',
+          links: {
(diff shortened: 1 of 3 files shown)


dpaez:
the `saveItem` function ends up creating a new hyperdrive and calling some internal methods like `stat` that may need to check the swarm and other peers. I recommend in this case to include the local dht to speed up things. The difference with only one test may not be huge but with others that made more operations could be significative.

One thing that adds some confusion and it's probably something I forgot to update in the past is the test description. From that description I imagine that this test should fail, I expect a throw, but reading it, it doesn't look to go that way. Unless, the invalid metadata refers to an empty main, but don't think so.

tests/test-save-item-method.js
```diff
@@ -0,0 +1,58 @@
...
+      description: 'lorem ipsum'
+    }
+
+    const { rawJSON } = await p2p.init(sampleData)
+
+    const key = rawJSON.url
+
+    try {
+      const { rawJSON: updated, metadata } = await p2p.saveItem({
+        isWritable: false,
```
dpaez, line 25: now I remember, I think this fails because the isWritable is set to false, hence the operation is invalid...

tests/test-save-item-method.js
```diff
@@ -0,0 +1,58 @@
...
+          p2pcommons: {
+            type: 'content',
+            subtype: '123',
+            main: '',
+            authors: [],
+            parents: []
+          }
+        }
+      })
+      t.same(updated.description, 'something new')
```
dpaez, line 50: But if the saveItem function fails, then it should throw and we dont reach this line.... 🤔

tests/test-save-item-method.js
```diff
@@ -0,0 +1,58 @@
...
+      t.same(metadata.version, 5)
+    } catch (err) {
+      t.fail(err)
```
dpaez, line 53: we expect an error from the test description

chartgerink:
Thanks @dpaez ☺️ I will close this issue then. I made a note of your comments for when we revisit the test.

PR #284, opened 2021-03-10
Add cancelable init method

### LINKED ISSUE #192: Make .clone() cancelable
**Is your feature request related to a problem? Please describe.**
Creating a module can take around ~20s sometimes. The user might choose to cancel creating the module. Currently the `.create` method isn't cancelable.

**Describe the solution you'd like**
Return a cancelable promise from `.create`.

**Describe alternatives you've considered**
Preventing the user from canceling..they will have to live with it 🤷

chartgerink:
This PR adds a cancelable init() method for the SDK. I am not 100% sure this is the create() function referred to in #192, but if so, closes #192. Otherwise, it also closes #192 because the create method no longer exists and it helps give clarity :blush:﻿

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,6 +155,7 @@ class SDK extends EventEmitter {
     // memoize methods
     this.ready = pMemoize(this.ready)
     // cancelable methods
+    this.init = PCancelable.fn(this.init.bind(this))
     this.clone = PCancelable.fn(this.clone.bind(this))
     // debug constructor
     debug(`platform: ${this.platform}`)


juliangruber:
There's more than this to it, check out the `clone()` method and how it has a `onCancel` argument. And you're right, this will close #192, I don't remember why it said `create()` but it should have been for `init()`

chartgerink:
> 
> 
> There's more than this to it, check out the `clone()` method and how it has a `onCancel` argument. And you're right, this will close #192, I don't remember why it said `create()` but it should have been for `init()`

Thanks @juliangruber --- I should've been a bit more dilligent here 😅 I will update with the relevant bits. Until then, this can be considered a draft.

juliangruber:
> Until then, this can be considered a draft.

You can convert this PR to a draft in the top right 🤟

chartgerink:
> 
> 
> > Until then, this can be considered a draft.
> 
> You can convert this PR to a draft in the top right 🤟

Damn, kept looking in the wrong places. Thank you 🧙

Issue #285, opened 2021-03-15
Clone method unreliable


chartgerink:
## Description

The clone method proves unreliable in cloning modules, both unversioned and versioned.

## To Reproduce

```js
const SDK = require('@p2pcommons/sdk-js')

const baseDir = '.test-sdk'
const x = new SDK({ baseDir })

const chris = "2ae99d1441b7605a9775b2e1c7835d4ddaaafc9472b0a0fd6e9bf9607dd210b5"
const hans = "e9ccc00b466a4d437f387e352f7098af5b022b8774a96852839a32ba19f47cc8"
const martin = "35c06f644725a71c24a2f83ed08489ee373b3f49dfdab44bbbca25549a67bdee"
const unversionedKey4 = "43eb22e34ad1f03615200dfcf16e092b8f571cc88b1355d0e9b77845b4052832"
const version4 = 25

;(async () => { 
    await x.ready()
    console.log('Started client')
    await x.clone(chris)
    console.log(`Successfully cloned ${chris}`)

    await x.clone(hans)
    console.log(`Successfully cloned ${hans}`)

    await x.clone(martin)
    console.log(`Successfully cloned ${martin}`)

    await x.clone(unversionedKey4, version4)
    console.log(`Successfully cloned ${unversionedKey4}+${version4}`)
})()

```

## Expected behavior

All modules available cloned successfully in one run.

## Actual behavior

Some modules get cloned on first run (`chris`), others on second (`hans`), others not at all (`martin` and checkout).

## System

- OS (incl. version): macOS 11.2.2
- SDK version:`main@ea8b6c12cefdfe5619d78be81d1756e130e4df4b`

## Additional context

All modules available from Hypegraph vault and permanent seeders

<details>
    <summary>First run output</summary>

```bash
chjh@module test-sdk-checkout % node index.js 
Started client
Successfully cloned 2ae99d1441b7605a9775b2e1c7835d4ddaaafc9472b0a0fd6e9bf9607dd210b5
(node:30500) UnhandledPromiseRejectionWarning: Error: clone: Problems fetching external module
    at SDK.getFromSwarm (/Users/chjh/playground/test-sdk-checkout/node_modules/@p2pcommons/sdk-js/index.js:1760:13)
[...]

PR #287, opened 2021-03-25
Update: alternative checkout fallback

dpaez:
This PR attempts to improve the situation described in #285 

It modifies the `getFromSwarm` function to try with the `latest` version if the version passed by the user is not available at the moment. This is done to avoid an issue with checkout specific copies and what is being seeded by the permanent seeder instances. Some drives are missing some blocks and this can block/hang any operation with a drive's checkout.  Whilst its clear that a complete fix may involve some work on the permanent seeder and drive completeness checks the current work in this PR improves the general situation and possible the UX.

The new fallback functionality is enabled by default in the clone method. You can turn it off by passing the option `latestFallback: false`. (internally is set to false to maintain the original behavior)


chartgerink (approved):
Thanks @dpaez! 

I tested this a bit more with Hypergraph, and adjusted some of the timeouts for the specific scenario. 

I was able to resolve the situation such that it doesn't need to rely on the fallback: A machine that has a p2pcommons instance with the specific version of a Hyperdrive can seed it to the network. As long as that machine is on (I have one running at this very moment for the dev version of the Vault `/discover` path) other instances have no issue retrieving the checkout. I still miss blocks if I try to replicate the same Hyperdrive using `hyperdrive-daemon`, but it's something.

This seems to indeed point towards that there is something about the seeding in the permanent seeder, and less of an issue with the SDK? We can discuss this sometime, maybe.

PR #290, opened 2021-05-07
Add import method

### LINKED ISSUE #212: Import p2pcommons method
A method to import p2pcommons for easy restore (linked to #211)

* the private keys
* `settings.json`

The method should:
- [ ] Take a path to a previously exported file (includes private keys and `settings.json`
- [ ] Import the private keys into a `~/.p2pcommons` folder
- [ ] Import the `settings.json` into a `~/.p2pcommons` folder
- [ ] Reconstruct p2pcommons database and content based on the information in `settings.json` (i.e., `defaultProfile` and `networkDepth`)

chartgerink:
This PR proposes an import method that synchronises all profiles upon import. Still some work to do before this can be considered final.

Closes #212.


index.js
```diff
@@ -1559,9 +1560,7 @@ class SDK extends EventEmitter {
       return archive
     }
 
-    const drive = dat.open(this.store, keyBuffer, {
-      sparse: true
-    })
+    const drive = dat.open(this.store, keyBuffer)
```
chartgerink, line 1563: For clarification and understanding: Why are we removing the sparse flag here?

index.js
```diff
@@ -2523,6 +2522,118 @@ class SDK extends EventEmitter {
...
+          }
+
+          this.assertMetadata(metadata)
+          await this.localdb.put(DatEncoding.encode(profile.rawJSON.url), {
+            ...metadata,
+            rawJSON: module,
+            avroType: this._getAvroType(module.p2pcommons.type).name
+          })
+        } else {
+          await this.clone(profile.rawJSON.url)
```
chartgerink, line 2631: Do I correctly understand this runs when the profile is not writable?

chartgerink (commented):
It's looking really good! Thanks @dpaez 🤗 

Left a few questions for understanding and starting points for our conversation tomorrow - looking forward to talking about next steps. 😊

index.js
dpaez: oh yes, this was just a little tweak. `dat.open` has a couple of default params, one is `sparse: true`, so it wasn't necessary here.

index.js
dpaez: Yes, you are right. Here is where we can adjust the `importBackup` method. First we are only considering profiles (before the if statement), maybe we can consider all modules? 
Then, I only focus on the writable profile, but again we can maybe consider all the writable modules. 

The other thing to take into account is what happens with non-writable profiles that are updated, maybe we would need to do something like we did with the writable profile. And simply remove the if/else part.

chartgerink:
Based on the conversation @dpaez and I had today, the following points are noted here for reference when we continue next week:

- Only writable profiles are the scope of `importBackup()`
- [ ] Set up a branch in Hypergraph that integrates this method (@chartgerink)
- [ ] Import with zero peers should fail (@dpaez)
- [ ] Assert internet connectivity (@dpaez)

Happy weekend!

index.js
```diff
@@ -2510,24 +2523,138 @@ class SDK extends EventEmitter {
...
+    debug('importBackup: starting...')
+    let zip
+    if (importPath.endsWith('.zip')) {
+      zip = new AdmZip(importPath)
+      if (!zip.getEntry('db/LOG')) throw new Error('no db')
+    } else {
+      throw new Error('No zip file supplied')
+    }
+
+    await this.assertOnline()
```
chartgerink, line 2547: This assertion doesn't throw when implemented in Hypergraph.

index.js
```diff
@@ -2510,24 +2523,138 @@ class SDK extends EventEmitter {
...
+    await zip.extractAllTo(this.baseDir)
+    // reopen session
+    await this.opendb()
+    await this.createStore()
+    await this.startSwarm()
+    await this.refreshMTimes()
+
+    await new Promise(resolve => setTimeout(resolve, 2500))
+
+    this.assertPeers()
```
chartgerink, line 2562: This assertion doesn't throw when implemented in Hypergraph (I think, not 100% on this one).

index.js
```diff
@@ -2510,24 +2523,138 @@ class SDK extends EventEmitter {
...
+
+          try {
+            await new Promise((resolve, reject) =>
+              pDrive.metadata.update({ ifAvailable: true }, err => {
+                if (err) return reject(err)
+                return resolve()
+              })
+            )
+          } catch (_) {
+            throw new Error('Unable to get Update')
```
chartgerink, line 2597: In my implementation in Hypergraph, I get the error that "Unable to get Update" error despite being able to get the drive from a peer in a fresh instance.

tests/test-export-backup-method.js
```diff
@@ -1,33 +1,157 @@
...
+      const { metadata } = await p2p.get(content.url)
+      const contentKeyVersion = `${encode(content.url)}+${metadata.version}`
+
+      await p2p.exportBackup(backupFile)
+      // Update the profile to check during import
+      await p2p.register(contentKeyVersion, encode(profile.url))
+      await p2p.refreshDrive(contentKeyVersion)
       // check that the database is correctly opened again
       await p2p.init(sampleContent)
       await p2p.listContent()
```
chartgerink, line 53: With the updates in the two lines above, this may prove redundant? What do you think?

tests/test-export-backup-method.js
```diff
@@ -1,33 +1,157 @@
...
+        })
+
+        const { rawJSON: updatedRawJSON } = await p2p3.get(content.rawJSON.url)
+
+        t.equal(
+          updatedRawJSON.description,
+          description,
+          'update after import works'
+        )
+
```
chartgerink, line 128: Maybe we should ensure that the other instance syncs again with the imported profile? This may help close the loop on the potential for conflicts on the network.

tests/test-export-backup-method.js
```diff
@@ -1,33 +1,157 @@
...
+      try {
+        // perform the import
+        await p2p4.importBackup(backupFile)
+        t.fail('Should throw ENOPEERS error')
+      } catch (err) {
+        t.ok(err instanceof SDK.errors.ENOPEERS, 'Expected ENOPEERS error')
+      } finally {
+        await p2p4.destroy()
+      }
+    })
```
chartgerink, line 155: Maybe we could include a test that simulates being offline? This situation is not included in the tests at the moment. I could not find any modules that do that though, so might not be a good idea hihi

chartgerink (commented):
Thanks @dpaez for the work that went into this!

I am sharing a few comments based on my experiences through implementing [this in Hypergraph](https://github.com/hypergraph-xyz/desktop/pull/449).

Looking forward to talking about this in a few hours!

chartgerink:
@dpaez and I talked about what is necessary to unlock real life testing with researchers (in order):

- [ ] Get the checks to pass (without tricks 😊; responsible: @dpaez)
- [ ] Ensure updates can be retrieved in practice (see also [Hypergraph implementation](https://github.com/hypergraph-xyz/desktop/pull/449) for testing) - responsible: ambiguous
- [ ] Ensure assertions for being online and peer availability throw in the Hypergraph implementation (responsible: @chartgerink)

We will reconvene on Friday to check in.

chartgerink:
Update on the assertions:

After upgrading to the latest version of this PR, the assertions throw as expected. See the dropdown below for more details

<details>
  <summary>Assertion testing</summary>
I started out by isolating whether the assertions throw as expected in the SDK itself. To do this, I created a quick script to test this

```js
const SDK = require('./index')

const baseDir = '.test-sdk'
const x = new SDK({ baseDir })

;(async () => { 
    await x.ready()
    console.log('Started client')

    await x.importBackup('/Users/chjh/Documents/hypergraph-key-before.zip')
    console.log('Imported backup')

})()
```

This threw as expected both for `EOFFLINE` and `ENOPEERS`. 

Subsequently I upgraded the branch for Hypergraph, and tested again. This indicated that I was catching the errors for the `importBackup()` method, which prevented it from throwing. After removing these in https://github.com/hypergraph-xyz/desktop/pull/449/commits/acd4f7e7e01b6f5eb4d4584bc00e05a5a59585e3 it all threw as expected (both `ENOPEERS` and `EOFFLINE`)
</details>

chartgerink:
A short reportback from the session @dpaez and I had 📞 

* @dpaez was able to get the checks to pass
* @chartgerink upgraded the implementation in Hypergraph to that commit (88f9bbb1258d4df487cecc2720dfa2f3819ba29e) after which the assertions threw as expected (more detailed description in previous comment)
* Together, we tested the `importBackup()` and whether it retrieves updates.
  * First, Hypergraph kept indicating there were no peers, when one was available on LAN
  * To isolate the issue, we tried it in the SDK only  with the following example snippet
<details>
<summary>Reproducible SDK import example (replace file)</summary>

```js
const SDK = require('./index')

const baseDir = '.test-sdk'
const x = new SDK({ baseDir })

;(async () => { 
    await x.ready()
    console.log('Started client')

    await x.importBackup('/Users/chjh/Documents/hypergraph-key-before.zip')
    console.log('Imported backup')

    console.log(await x.listProfiles())

})()
```

</details>

* The `importBackup()` successfully updated and restored the SDK instance
* This helped us realise the issue was not with the SDK, and that something had to be different in the implementation in Hypergraph
* After some digging, we discovered the `package-lock.json` had an unexpected difference
  *  `@p2pcommons/sdk-js`: `hypercore@9.5.6`
  * Hypergraph: `hypercore@9.10.0` (latest)
* When we forced `npm install hypercore@9.5.6` for Hypergraph, the restore worked, indicating there is some as of yet unknown difference between these versions.

All in all: `importBackup()` works! 🥳 

## Next steps

- [ ] @dpaez will fix the dependency `hypercore@9.5.6` for the SDK
- [ ] @dpaez will create a reproducible example of the regressive issue for hypercore
- [ ] @chartgerink will review the PR one last time
- [ ] @chartgerink will continue testing in Hypergraph and further develop that UX
- [ ] We decide whether the PR is merged before or after testing with researchers, together

.github/workflows/ci.yml
```diff
@@ -5,12 +5,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 13.x]
+        node-version: [10.x, 12.x, 14.x]
```
chartgerink, line 8: Why are we dropping version 13 specifically? I can imagine shifting everything up by one, and want to make sure I understand this decision.

chartgerink (commented):
Haha I can't approve the PR - but looks good. 

Only have one question but I don't think that affects our potential to merge on its own.

.github/workflows/ci.yml
dpaez: Just to use the LTS versions of node. Even-numbered are experimental versions, good to test new features coming to the next stable version. But we can focus on LTS and that would be fine.

Here it is the node releases plan: https://nodejs.org/en/about/releases/ 
Based on that, node 10 is almost vanishing, and in October  node 16 would become LTS, and then we can safely switch to `[12, 14, 16]`.

chartgerink:
Update as I return from my break: This is still being tested in Hypergraph Desktop, and is not a stale PR. 😊

PR #293, opened 2021-06-06
Update: increase startup related timers

dpaez:
This PR increases some timers used on the sdk startup. If they timeout too early this can lead to skip some re-seeding calls and hence omitting the announcing of some drives. 

This can be related to https://github.com/geut/permanent-seeder/issues/164


chartgerink (approved):
Seems good 😊

Are the upgrades to the dependency maintenance or also relate to the startup timers? I expect not and want to check my assumption. 

Regardless, looks like there are a few changes that seem like good maintenance, and the increased timeouts look reasonable - looking at the test times in the checks also doesn't create increases in the amount of time it takes to run then (average test runtime new v old: 45.6 sec v 44 sec)

dpaez:
> Are the upgrades to the dependency maintenance or also relate to the startup timers? I expect not and want to check my assumption.

Yes, you are right. They are not related. Since I was checking the checkout issue at the same time I thought it was a good idea to update deps in the meantime.

